PR #39, opened 2015-03-16
Clee/ma 77 auto course reruns

BenjiLee:
@nasthagiri Who else should I tag?

Going to combine the tickets. You can start the review on fad19b7.

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -278,11 +278,9 @@ def get_videos_for_ids(
     return (VideoSerializer(video).data for video in videos)
 
 
-def get_video_info_for_course_and_profile(course_id, profile_name):
+def get_video_info_for_course_and_profiles(course_id, profiles):
     """
     Returns a dict of edx_video_ids with a dict of requested profiles.
-
-    If the profiles or video is not found, urls will be blank.
 
     Args:
         course_id (str): id of the course
@@ -292,48 +290,51 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
         edx_video_id
         {
             edx_video_id: {
-                profile_name: {
-                    'url': url of the encoding
-                    'duration': length of the video in seconds
-                    'file_size': size of the file in bytes
-                },
+                'duration': length of the video in seconds,
+                'profiles': {
+                    profile_name: {
+                        'url': url of the encoding
+                        'file_size': size of the file in bytes
+                    },
+                }
             },
         }
     Example:
         Given two videos with two profiles each in course_id 'test_course':
         {
             u'edx_video_id_1': {
-                u'mobile': {
-                    'url': u'http: //www.example.com/meow',
-                    'duration': 1111,
-                    'file_size': 2222
-                },
-                u'desktop': {
-                    'url': u'http: //www.example.com/woof',
-                    'duration': 3333,
-                    'file_size': 4444
+                u'duration: 1111,
+                u'profiles': {
+                    u'mobile': {
+                        'url': u'http: //www.example.com/meow',
+                        'file_size': 2222
+                    },
+                    u'desktop': {
+                        'url': u'http: //www.example.com/woof',
+                        'file_size': 4444
+                    }
                 }
             },
             u'edx_video_id_2': {
-                u'mobile': {
-                    'url': u'http: //www.example.com/roar',
-                    'duration': 5555,
-                    'file_size': 6666
-                },
-                u'desktop': {
-                    'url': u'http: //www.example.com/bzzz',
-                    'duration': 7777,
-                    'file_size': 8888
+                u'duration: 2222,
+                u'profiles': {
+                    u'mobile': {
+                        'url': u'http: //www.example.com/roar',
+                        'file_size': 6666
+                    },
+                    u'desktop': {
+                        'url': u'http: //www.example.com/bzzz',
+                        'file_size': 8888
+                    }
                 }
             }
         }
     """
-    #TODO This function needs unit tests. Write them when addressing MA-169
     # In case someone passes in a key (VAL doesn't really understand opaque keys)
     course_id = unicode(course_id)
     try:
         encoded_videos = EncodedVideo.objects.filter(
-            profile__profile_name=profile_name,
+            profile__profile_name__in=profiles,
             video__courses__course_id=course_id
         ).select_related()
     except Exception:
@@ -342,36 +343,39 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
         raise ValInternalError(error_message)
 
     # DRF serializers were causing extra queries for some reason...
-    return {
-        enc_vid.video.edx_video_id: {
-            "url": enc_vid.url,
-            "file_size": enc_vid.file_size,
-            "duration": enc_vid.video.duration,
-        }
-        for enc_vid in encoded_videos
-    }
+    return_dict = {}
+    for enc_vid in encoded_videos:
+        # Add duration to edx_video_id
+        return_dict.setdefault(enc_vid.video.edx_video_id, {}).update(
+            {
+                "duration": enc_vid.video.duration,
+            }
+        )
+        # Add profile information to edx_video_id's profiles
+        return_dict[enc_vid.video.edx_video_id].setdefault("profiles", {}).update(
+            {enc_vid.profile.profile_name: {
+                "url": enc_vid.url,
+                "file_size": enc_vid.file_size,
+            }}
+        )
+    return return_dict
 
 
-def copy_course_videos(old_course_id, new_course_id):
+def copy_course_videos(source_course_id, destination_course_id):
     """
-    Adds the new_course_id to the videos taken from the old_course_id
+    Adds the destination_course_id to the videos taken from the source_course_id
 
     Args:
-        old_course_id: The original course_id
-        new_course_id: The new course_id of the rerun
+        source_course_id: The original course_id
+        destination_course_id: The new course_id where the videos will be copied
     """
-    if old_course_id == new_course_id:
-        raise ValueError(
-            "Both course_id's are the same: {}".format(new_course_id)
-        )
-    if Video.objects.filter(courses__course_id=unicode(new_course_id)):
-        raise ValCannotCreateError(
-            "New course id already exists: {}".format(new_course_id)
-        )
-    videos = Video.objects.filter(courses__course_id=unicode(old_course_id))
+    if source_course_id == destination_course_id:
+        return
+
+    videos = Video.objects.filter(courses__course_id=unicode(source_course_id))
 
     for video in videos:
-        CourseVideo.objects.create(
+        CourseVideo.objects.get_or_create(
             video=video,
-            course_id=new_course_id
+            course_id=destination_course_id
         )
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -6,7 +6,7 @@ The internal API for VAL. This is not yet stable
 from enum import Enum
 import logging
 
-from edxval.models import Video, EncodedVideo
+from edxval.models import Video, EncodedVideo, CourseVideo
 from edxval.serializers import VideoSerializer, ProfileSerializer
 
 logger = logging.getLogger(__name__)  # pylint: disable=C0103
@@ -93,6 +93,7 @@ def create_video(video_data):
                         extension: 3 letter extension of video
                         width: horizontal pixel resolution
                         height: vertical pixel resolution
+                courses: Courses associated with this video
          }
     """
     serializer = VideoSerializer(data=video_data)
@@ -119,7 +120,7 @@ def create_profile(profile_data):
         }
 
     Returns:
-        new_object.id (int): id of the newly created object
+        (int): id of the newly created object
 
     Raises:
         ValCannotCreateError: Raised if the serializer throws an error
@@ -141,7 +142,7 @@ def get_video_info(edx_video_id, location=None):  # pylint: disable=W0613
         edx_video_id (str): id for video content.
 
     Returns:
-        result (dict): Deserialized Video Object with related field EncodedVideo
+        (dict): Deserialized Video Object with related field EncodedVideo
             Returns all the Video object fields, and it's related EncodedVideo
             objects in a list.
             {
@@ -180,7 +181,7 @@ def get_video_info(edx_video_id, location=None):  # pylint: disable=W0613
             'client_video_id': u'The example video',
             'encoded_videos': [
                 {
-                    'url': u'http://www.meowmix.com',
+                    'url': u'http://www.example.com',
                     'file_size': 25556,
                     'bitrate': 9600,
                     'profile': u'mobile'
@@ -212,7 +213,7 @@ def get_urls_for_profiles(edx_video_id, profiles):
         profiles (list): list of profiles we want to search for
 
     Returns:
-        profiles_to_urls (dict): A dict containing the profile to url pair
+        (dict): A dict containing the profile to url pair
     """
     profiles_to_urls = {profile: None for profile in profiles}
     try:
@@ -236,7 +237,7 @@ def get_url_for_profile(edx_video_id, profile):
         profile (str): a string of the profile we are searching
 
     Returns:
-        A string with the url
+        (str): A string with the url
 
     """
     return get_urls_for_profiles(edx_video_id, [profile])[profile]
@@ -279,10 +280,55 @@ def get_videos_for_ids(
 
 def get_video_info_for_course_and_profile(course_id, profile_name):
     """
-    Returns a dict mapping profiles to URLs.
+    Returns a dict of edx_video_ids with a dict of requested profiles.
 
     If the profiles or video is not found, urls will be blank.
+
+    Args:
+        course_id (str): id of the course
+        profiles (list): list of profile_names
+    Returns:
+        (dict): Returns all the profiles attached to a specific
+        edx_video_id
+        {
+            edx_video_id: {
+                profile_name: {
+                    'url': url of the encoding
+                    'duration': length of the video in seconds
+                    'file_size': size of the file in bytes
+                },
+            },
+        }
+    Example:
+        Given two videos with two profiles each in course_id 'test_course':
+        {
+            u'edx_video_id_1': {
+                u'mobile': {
+                    'url': u'http: //www.example.com/meow',
+                    'duration': 1111,
+                    'file_size': 2222
+                },
+                u'desktop': {
+                    'url': u'http: //www.example.com/woof',
+                    'duration': 3333,
+                    'file_size': 4444
+                }
+            },
+            u'edx_video_id_2': {
+                u'mobile': {
+                    'url': u'http: //www.example.com/roar',
+                    'duration': 5555,
+                    'file_size': 6666
+                },
+                u'desktop': {
+                    'url': u'http: //www.example.com/bzzz',
+                    'duration': 7777,
+                    'file_size': 8888
+                }
+            }
+        }
     """
+    #TODO This function needs unit tests. Write them when addressing MA-169
     # In case someone passes in a key (VAL doesn't really understand opaque keys)
     course_id = unicode(course_id)
     try:
@@ -304,3 +350,28 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
         }
         for enc_vid in encoded_videos
     }
+
+
+def copy_course_videos(old_course_id, new_course_id):
+    """
+    Adds the new_course_id to the videos taken from the old_course_id
+
+    Args:
+        old_course_id: The original course_id
+        new_course_id: The new course_id of the rerun
+    """
+    if old_course_id == new_course_id:
+        raise ValueError(
+            "Both course_id's are the same: {}".format(new_course_id)
+        )
+    if Video.objects.filter(courses__course_id=unicode(new_course_id)):
+        raise ValCannotCreateError(
+            "New course id already exists: {}".format(new_course_id)
+        )
+    videos = Video.objects.filter(courses__course_id=unicode(old_course_id))
+
+    for video in videos:
+        CourseVideo.objects.create(
+            video=video,
+            course_id=new_course_id
+        )
--- a/edxval/tests/constants.py
+++ b/edxval/tests/constants.py
@@ -30,13 +30,33 @@ Encoded_videos for test_api, does not have profile.
 """
 ENCODED_VIDEO_DICT_MOBILE = dict(
     url="http://www.meowmix.com",
-    file_size=4545,
-    bitrate=6767,
+    file_size=11,
+    bitrate=22,
 )
 ENCODED_VIDEO_DICT_DESKTOP = dict(
     url="http://www.meowmagic.com",
-    file_size=1212,
-    bitrate=2323,
+    file_size=33,
+    bitrate=44,
+)
+ENCODED_VIDEO_DICT_MOBILE2 = dict(
+    url="http://www.woof.com",
+    file_size=55,
+    bitrate=66,
+)
+ENCODED_VIDEO_DICT_DESKTOP2 = dict(
+    url="http://www.bark.com",
+    file_size=77,
+    bitrate=88,
+)
+ENCODED_VIDEO_DICT_MOBILE3 = dict(
+    url="http://www.ssss.com",
+    file_size=1111,
+    bitrate=2222,
+)
+ENCODED_VIDEO_DICT_DESKTOP3 = dict(
+    url="http://www.hiss.com",
+    file_size=3333,
+    bitrate=4444,
 )
 """
 Validators
@@ -322,6 +342,18 @@ COMPLETE_SET_EXTRA_VIDEO_FIELD = dict(
 """
 Other
 """
+VIDEO_DICT_TREE = dict(
+    client_video_id="Trees4lyfe",
+    duration=532.00,
+    edx_video_id="tree-hugger",
+    status="test",
+)
+VIDEO_DICT_PLANT = dict(
+    client_video_id="PlantzRule",
+    duration=876.00,
+    edx_video_id="fernmaster",
+    status="test",
+)
 VIDEO_DICT_ZEBRA = dict(
     client_video_id="Zesty Zebra",
     duration=111.00,
--- a/edxval/tests/test_api.py
+++ b/edxval/tests/test_api.py
@@ -14,9 +14,14 @@ from ddt import ddt, data
 
 from edxval.models import Profile, Video, EncodedVideo, CourseVideo
 from edxval import api as api
-from edxval.api import SortDirection, ValCannotCreateError, VideoSortField
+from edxval.api import (
+    SortDirection,
+    ValCannotCreateError,
+    VideoSortField,
+)
 from edxval.serializers import VideoSerializer
 from edxval.tests import constants, APIAuthTestCase
+
 
 @ddt
 class CreateVideoTest(TestCase):
@@ -125,16 +130,12 @@ class GetVideoInfoTest(TestCase):
         Profile.objects.create(**constants.PROFILE_DICT_DESKTOP)
         video = Video.objects.create(**constants.VIDEO_DICT_FISH)
         EncodedVideo.objects.create(
-            video=Video.objects.get(
-                edx_video_id=constants.VIDEO_DICT_FISH.get("edx_video_id")
-            ),
+            video=video,
             profile=Profile.objects.get(profile_name="mobile"),
             **constants.ENCODED_VIDEO_DICT_MOBILE
         )
         EncodedVideo.objects.create(
-            video=Video.objects.get(
-                edx_video_id=constants.VIDEO_DICT_FISH.get("edx_video_id")
-            ),
+            video=video,
             profile=Profile.objects.get(profile_name="desktop"),
             **constants.ENCODED_VIDEO_DICT_DESKTOP
         )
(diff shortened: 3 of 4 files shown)


BenjiLee:
@nasthagiri b5d8b07 is also ready for review. I will post the related server side changes here.

edxval/api.py
```diff
@@ -277,17 +278,63 @@ def get_videos_for_ids(
     return (VideoSerializer(video).data for video in videos)
 
 
-def get_video_info_for_course_and_profile(course_id, profile_name):
+def get_video_info_for_course_and_profiles(course_id, profiles):
     """
-    Returns a dict mapping profiles to URLs.
+    #TODO Change this to reflect new changes
```
nasthagiri: Is this TODO still needed?

edxval/api.py
```diff
@@ -296,11 +343,47 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+    if Video.objects.filter(courses__course_id=unicode(new_course_id)):
+        raise ValCannotCreateError(
+            "New course id already exists: {}".format(new_course_id)
+        )
+    videos = Video.objects.filter(courses__course_id=unicode(old_course_id))
+
+    if len(videos) == 0:
+        raise ValVideoNotFoundError(
+            "Cannot find any videos for given course_id: {}".format(old_course_id)
+        )
```
nasthagiri: I don't think this error is needed.  We should support having courses without any videos.  We also don't want to require the caller to first check if the original course has gone through the video-pipeline process and has videos in VAL before calling this API.

edxval/tests/test_api.py
```diff
@@ -436,3 +530,90 @@ def test_get_info_queries_for_only_video(self):
         with self.assertNumQueries(4):
             api.get_video_info(constants.VIDEO_DICT_ZEBRA.get("edx_video_id"))
 
+@ddt
+class RerunCourse(TestCase):
```
nasthagiri: Rename the class to `TestRerunCourse`?

BenjiLee:
@gwprice 

The other side of the world. https://github.com/edx/edx-platform/pull/7397

edxval/api.py
```diff
@@ -277,17 +278,62 @@ def get_videos_for_ids(
...
-    Returns a dict mapping profiles to URLs.
+    Returns a dict of edx_video_ids with a dict of profiles.
 
     If the profiles or video is not found, urls will be blank.
+
+    Args:
+        course_id (str): id of the course
+        profiles (list): list of profile_names
+    Returns:
+        return_dict (dict): Returns all the profiles attached to a specific
```
gwprice: s/return_dict //

edxval/api.py
```diff
@@ -277,17 +278,62 @@ def get_videos_for_ids(
...
+                },
+            },
+        }
+    Example:
+        Given two videos with two profiles each in course_id 'test_course':
+        >>> get_video_info_for_course_and_profiles('test-course', ['mobile', 'desktop''])
+        {
+            u'super-soaker': {
+                u'mobile': {
+                    'url': u'http: //www.meowmix.com',
```
gwprice: Use example.com for example URLs. Also, the example should distinguish between the URLs for the two videos. I suggest something like `http://example.com/super-soaker-mobile.mp4`

edxval/api.py
```diff
@@ -296,11 +342,47 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+def rerun_course(old_course_id, new_course_id):
+    """
+    Adds the new_course_id to the videos taken from the old_course_id
+
+    Args:
+        old_course_id: The original course_id
+        new_course_id: The new course_id of the rerun
+    """
+    if old_course_id == new_course_id:
+        raise ValCannotCreateError(
```
gwprice: Just use `ValueError`, which is Python's general error for invalid arguments

edxval/api.py
```diff
@@ -296,11 +342,47 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+                    "file_size": enc_vid.file_size,
+                    "duration": enc_vid.video.duration,
+                }
+            }
+        )
+
+    return return_dict
+
+
+def rerun_course(old_course_id, new_course_id):
```
gwprice: I think a more generic name like `copy_course_videos` would be better

edxval/api.py
```diff
@@ -296,11 +342,47 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+
+    Args:
+        old_course_id: The original course_id
+        new_course_id: The new course_id of the rerun
+    """
+    if old_course_id == new_course_id:
+        raise ValCannotCreateError(
+            "Both course_id's are the same: {}".format(new_course_id)
+        )
+    if Video.objects.filter(courses__course_id=unicode(new_course_id)):
```
gwprice: I don't think this should be considered an error. The onus should be on the caller to be certain that they want to copy the videos, and I don't see a compelling reason to disallow this case.

edxval/api.py
```diff
@@ -296,11 +342,47 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+        )
+    videos = Video.objects.filter(courses__course_id=unicode(old_course_id))
+
+    if len(videos) == 0:
+        raise ValVideoNotFoundError(
+            "Cannot find any videos for given course_id: {}".format(old_course_id)
+        )
+
+    for video in videos:
+        CourseVideo.objects.create(
```
gwprice: If the restriction against calling this function for a pre-existing course is removed, this should change to `get_or_create`

edxval/api.py
gwprice: Alternatively, this could just be considered a no-op instead of an error

edxval/tests/test_api.py
```diff
@@ -269,6 +275,94 @@ def test_get_url_for_profile(self):
...
+        """
+        Profile.objects.create(**constants.PROFILE_DICT_MOBILE)
+        Profile.objects.create(**constants.PROFILE_DICT_DESKTOP)
+        # 1st video
+        video = Video.objects.create(**constants.VIDEO_DICT_FISH)
+        EncodedVideo.objects.create(
+            video=Video.objects.get(
+                edx_video_id=constants.VIDEO_DICT_FISH.get("edx_video_id")
+            ),
+            profile=Profile.objects.get(profile_name="mobile"),
```
gwprice: Save the result of `create` above to a variable and use it here instead of doing a get

edxval/tests/test_api.py
```diff
@@ -269,6 +275,94 @@ def test_get_url_for_profile(self):
...
+
+        Creates two videos, each with two encodings.
+        """
+        Profile.objects.create(**constants.PROFILE_DICT_MOBILE)
+        Profile.objects.create(**constants.PROFILE_DICT_DESKTOP)
+        # 1st video
+        video = Video.objects.create(**constants.VIDEO_DICT_FISH)
+        EncodedVideo.objects.create(
+            video=Video.objects.get(
+                edx_video_id=constants.VIDEO_DICT_FISH.get("edx_video_id")
```
gwprice: The "edx_video_id" key is guaranteed to be there, so use brackets instead of get

gwprice: Never mind, this is about to be unnecessary

edxval/api.py
BenjiLee: :) Was struggling to find a name for this.

edxval/tests/test_api.py
```diff
@@ -269,6 +275,94 @@ def test_get_url_for_profile(self):
...
+        Creates a course for testing
+
+        Creates two videos, each with two encodings.
+        """
+        Profile.objects.create(**constants.PROFILE_DICT_MOBILE)
+        Profile.objects.create(**constants.PROFILE_DICT_DESKTOP)
+        # 1st video
+        video = Video.objects.create(**constants.VIDEO_DICT_FISH)
+        EncodedVideo.objects.create(
+            video=Video.objects.get(
```
gwprice: Just use the video variable

edxval/api.py
```diff
@@ -277,17 +278,62 @@ def get_videos_for_ids(
     return (VideoSerializer(video).data for video in videos)
 
 
-def get_video_info_for_course_and_profile(course_id, profile_name):
+def get_video_info_for_course_and_profiles(course_id, profiles):
     """
-    Returns a dict mapping profiles to URLs.
+    Returns a dict of edx_video_ids with a dict of profiles.
 
     If the profiles or video is not found, urls will be blank.
```
gwprice: This seems to be out of date. I think the description should explicitly state that a video without an encoding in one of the specified profiles will not be in the result.

edxval/tests/test_api.py
```diff
@@ -269,6 +275,94 @@ def test_get_url_for_profile(self):
...
 
 
+class GetVideoForCourseProfiles(TestCase):
+    """Tests get_video_info_for_course_and_profiles in api.py"""
+
+    def setUp(self):
+        """
+        Creates a course for testing
+
+        Creates two videos, each with two encodings.
```
gwprice: The test would be a bit stronger if if had a video in another course (to ensure that it gets filtered out)

edxval/tests/test_api.py
```diff
@@ -269,6 +275,94 @@ def test_get_url_for_profile(self):
...
+    def test_get_video_for_course_profiles_repeated_profile(self):
+        """
+        Tests get_video_info_for_course_and_profiles with repeated profile
+        """
+        videos = api.get_video_info_for_course_and_profiles(
+            'test-course',
+            ['mobile', 'mobile'])
+        self.assertEqual(len(videos[constants.VIDEO_DICT_FISH.get("edx_video_id")]), 1)
+        self.assertEqual(len(videos[constants.VIDEO_DICT_STAR.get("edx_video_id")]), 1)
+
```
gwprice: Too many blank lines here. Also, add test cases for specifying just one of the two profiles and for specifying an empty list of profiles

gwprice: Ah, I now see that the single profile case is covered; I think it would make sense to split the _success method into multiple cases

edxval/tests/test_api.py
```diff
@@ -269,6 +275,94 @@ def test_get_url_for_profile(self):
...
+        CourseVideo.objects.create(video=video, course_id=self.course_id)
+
+    def test_get_video_for_course_profiles_success(self):
+        """
+        Tests series of successful get_video_info_for_course_and_profiles calls
+        """
+        videos = api.get_video_info_for_course_and_profiles(
+            'test-course',
+            ['mobile'])
+        self.assertEqual(len(videos[constants.VIDEO_DICT_FISH.get("edx_video_id")]), 1)
```
gwprice: There should be a stronger assertion here about the contents of the result.

```
self.assertEqual(
    videos,
    {
        video_dict["edx_video_id"]: {
            PROFILE_DICT_MOBILE["profile_name"]: {
                "url": ENCODED_VIDEO_DICT_MOBILE["url"],
                "file_size": ENCODED_VIDEO_DICT_MOBILE["file_size"],
                "duration": video_dict["duration"],
            }
        }
        for video_dict in [VIDEO_DICT_FISH, VIDEO_DICT_STAR]
    }
)
```

edxval/tests/test_api.py
gwprice: There should also be a video that has an encoding in one profile but not the other.

edxval/tests/test_api.py
```diff
@@ -436,3 +530,90 @@ def test_get_info_queries_for_only_video(self):
...
 
+@ddt
+class TestRerunCourse(TestCase):
+    """Tests rerun course"""
+
+    def setUp(self):
+        """
+        Creates a course for testing
+
+        Creates two videos, each with two encodings.
```
gwprice: EncodedVideos are unnecessary for this test

gwprice: This should also create a video not in the test course to make sure that only videos in the test course get copied

edxval/tests/test_api.py
```diff
@@ -436,3 +530,90 @@ def test_get_info_queries_for_only_video(self):
...
+        """
+        courses = VideoSerializer(video).data['courses']
+        self.assertEqual(sorted(courses), sorted(expected_course_list))
+
+    def test_successful_rerun(self):
+        """Tests a successful rerun and 2nd rerun"""
+        api.rerun_course('test-course', 'meow-rerun')
+        videos = Video.objects.all()
+        for video in videos:
+            self._check_video_course_list(video,
```
gwprice: Fix whitespace here

edxval/api.py
```diff
@@ -296,11 +341,41 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+        new_course_id: The new course_id of the rerun
+    """
+    if old_course_id == new_course_id:
+        raise ValueError(
+            "Both course_id's are the same: {}".format(new_course_id)
+        )
+    if Video.objects.filter(courses__course_id=unicode(new_course_id)):
+        raise ValCannotCreateError(
+            "New course id already exists: {}".format(new_course_id)
+        )
```
BenjiLee: I think there was a comment here that got deleted. @gwprice Did you say that we should allow both of the cases     `New course id already exists: {}".format(new_course_id)` and `"Both course_id's are the same: {}".format(new_course_id)`?

gwprice: Yes, I think we should allow the course ids to be the same and treat it as a no-op and allow target courses that already have videos. The onus should be on the caller to make sure that's what they really want.

edxval/api.py
```diff
@@ -296,11 +341,41 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
-        for enc_vid in encoded_videos
-    }
+    return_dict = {}
+    for enc_vid in encoded_videos:
+        return_dict.setdefault(enc_vid.video.edx_video_id, {}).update(
+            {
+                enc_vid.profile.profile_name: {
+                    "url": enc_vid.url,
+                    "file_size": enc_vid.file_size,
+                    "duration": enc_vid.video.duration,
```
gwprice: It's a little weird to include the duration with each encoding, when it's a property of the video and not the encoding. The fact that the sole intended caller of this ends up filtering out the duration suggests that we should consider modifying the returned object to something like the following:

```
{
    "video_id_1": {
        "duration": 100,
        "encodings": {
            "profile1": {
                "url": "http://example.com/foo.mp4",
                "file_size": 555,
            },
            ...
        }
    },
    ...
```

edxval/api.py
```diff
@@ -296,11 +341,36 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+                    "url": enc_vid.url,
+                    "file_size": enc_vid.file_size,
+                }
+            }
+        )
+
+    return return_dict
+
+
+def copy_course_videos(old_course_id, new_course_id):
```
nasthagiri: Is this method idempotent?  If you call it again with the same argument values, it will throw an error, right?

edxval/tests/test_api.py
```diff
@@ -436,3 +620,78 @@ def test_get_info_queries_for_only_video(self):
...
+        """Tests when the new course id name is the same as the original"""
+        # with self.assertRaises(ValueError):
+        #     api.copy_course_videos('test-course', 'test-course')
+        api.copy_course_videos('test-course', 'test-course')
+
+    def test_existing_new_course_id_for_rerun(self):
+        """Test when the new course id already exists"""
+        api.copy_course_videos('test-course', 'meow-rerun')
+        # with self.assertRaises(ValCannotCreateError):
+        #     api.copy_course_videos('test-course', 'meow-rerun')
```
nasthagiri: should remove commented out code.

edxval/api.py
```diff
@@ -296,11 +341,36 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+        old_course_id: The original course_id
+        new_course_id: The new course_id of the rerun
+    """
+    if old_course_id == new_course_id:
+        return
+
+    videos = Video.objects.filter(courses__course_id=unicode(old_course_id))
+
+    for video in videos:
+        CourseVideo.objects.create(
```
nasthagiri: this should be changed to `get_or_create` since you are allowing videos with the `new_course_id` to already exist.

edxval/api.py
```diff
@@ -296,11 +341,36 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
-            "duration": enc_vid.video.duration,
-        }
-        for enc_vid in encoded_videos
-    }
+    return_dict = {}
+    for enc_vid in encoded_videos:
+        return_dict.setdefault(enc_vid.video.edx_video_id, {}).update(
+            {
+                "duration": enc_vid.video.duration,
+                enc_vid.profile.profile_name: {
```
gwprice: This needs to be embedded in another dict to avoid conflicts between a profile name and "duration"

BenjiLee: I'm not sure I follow.

gwprice: If someone creates a profile with the name "duration", this will break. As per the original comment, I think it should be like the following:

```
{
    "video_id_1": {
        "duration": 100,
        "encodings": {
            "profile1": {
                "url": "http://example.com/foo.mp4",
                "file_size": 555,
            },
            ...
        }
    },
    ...
}
```

edxval/api.py
gwprice: It is now idempotent.

nasthagiri: Yes, it is now.  When I had added that comment, the code was using `get` instead of `get_or_create`.  lgtm.

edxval/tests/test_api.py
```diff
@@ -436,3 +622,74 @@ def test_get_info_queries_for_only_video(self):
         with self.assertNumQueries(4):
             api.get_video_info(constants.VIDEO_DICT_ZEBRA.get("edx_video_id"))
 
+
+class TestRerunCourse(TestCase):
```
nasthagiri: Should be renamed to `TestCopyCourse`.

edxval/tests/test_serializers.py
```diff
@@ -1,123 +1,123 @@
...
-        """
-        Tests an non dict input in the VideoSerializer
-        """
-        data = "hello"
-        serializer = VideoSerializer(data=data)
-        self.assertEqual(
-            serializer.errors.get("non_field_errors")[0],
-            "Invalid data"
-        )
+# # pylint: disable=E1101
```
nasthagiri: Need to uncomment these tests.

BenjiLee:
@nasthagiri @gwprice

edxval/api.py
```diff
@@ -296,11 +339,39 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+    return return_dict
+
+
+def copy_course_videos(old_course_id, new_course_id):
+    """
+    Adds the new_course_id to the videos taken from the old_course_id
+
+    Args:
+        old_course_id: The original course_id
+        new_course_id: The new course_id of the rerun
```
nasthagiri: nitpick: I wouldn't say "rerun" here since this method is useful without it needing to be for the rerun feature.

edxval/tests/constants.py
```diff
@@ -30,13 +30,33 @@
...
+)
+ENCODED_VIDEO_DICT_MOBILE3 = dict(
+    url="http://www.ssss.com",
+    file_size=1111,
+    bitrate=2222,
+)
+ENCODED_VIDEO_DICT_DESKTOP3 = dict(
+    url="http://www.hiss.com",
+    file_size=3333,
+    bitrate=4444,
```
nasthagiri: Nitpick: you don't necessarily need a  new encoded_video_dict for each encoding.  The same one can be reused, no? Otherwise, it adds extra baggage to the test code.

edxval/tests/test_api.py
```diff
@@ -269,6 +270,191 @@ def test_get_url_for_profile(self):
...
+                }}
+            )
+        return video_dict
+
+    def test_get_video_for_course_profiles_success_one_profile(self):
+        """
+        Tests get_video_info_for_course_and_profiles for one profile
+        """
+        videos = api.get_video_info_for_course_and_profiles(
+            'test-course',
```
nasthagiri: Nitpick: reuse `self.course_id` here.

edxval/tests/test_api.py
```diff
@@ -436,3 +622,74 @@ def test_get_info_queries_for_only_video(self):
...
+        videos = Video.objects.all()
+        for video in videos:
+            self._check_video_course_list(video,
+                                          ['test-course',
+                                           'meow-rerun',
+                                           'woof-bark'])
+
+    def test_same_course_ids(self):
+        """Tests when the new course id name is the same as the original"""
+        api.copy_course_videos('test-course', 'test-course')
```
nasthagiri: need to add assert here to make sure the list of courses for the video remains the same.

edxval/tests/test_api.py
```diff
@@ -436,3 +622,74 @@ def test_get_info_queries_for_only_video(self):
...
+
+    def test_existing_new_course_id_for_rerun(self):
+        """Test when the new course id already exists"""
+        api.copy_course_videos('test-course', 'meow-rerun')
```
nasthagiri: ?  where does the `meow-rerun` course-id already exist for the video?
Once fixed, need to add assert here to make sure the list of courses for the video remains the same.

edxval/api.py
```diff
@@ -277,17 +278,59 @@ def get_videos_for_ids(
     return (VideoSerializer(video).data for video in videos)
 
 
-def get_video_info_for_course_and_profile(course_id, profile_name):
+def get_video_info_for_course_and_profiles(course_id, profiles):
     """
-    Returns a dict mapping profiles to URLs.
+    Returns a dict of edx_video_ids with a dict of requested profiles.
 
     If the profiles or video is not found, urls will be blank.
```
gwprice: This comment doesn't make sense; if a profile is missing, it will simply not be present. We should also note that a video with no encodings will not be included in the result.

edxval/api.py
```diff
@@ -277,17 +278,59 @@ def get_videos_for_ids(
...
+    Args:
+        course_id (str): id of the course
+        profiles (list): list of profile_names
+    Returns:
+        (dict): Returns all the profiles attached to a specific
+        edx_video_id
+        {
+            edx_video_id: {
+                'duration': length of the video in seconds,
+                profile_name: {
```
gwprice: This needs to be updated (as does the example below)

edxval/tests/test_api.py
```diff
@@ -269,6 +270,191 @@ def test_get_url_for_profile(self):
...
+        # 4th video different course
+        video = Video.objects.create(**constants.VIDEO_DICT_PLANT)
+        EncodedVideo.objects.create(
+            video=video,
+            profile=desktop_profile,
+            **constants.ENCODED_VIDEO_DICT_DESKTOP3
+        )
+        CourseVideo.objects.create(video=video, course_id=self.course_id2)
+
+    def _create_video_dict(self, video, profiles, encoded_videos):
```
gwprice: Calls to this would be clearer if it took a dictionary from profile name to encoded video instead taking profiles and encoded_videos separately as lists.

edxval/tests/test_api.py
gwprice: Reminder: The latter two of these have not been addressed yet. (GitHub seems to have moved the first two from elsewhere.)

edxval/tests/test_api.py
```diff
@@ -436,3 +622,74 @@ def test_get_info_queries_for_only_video(self):
...
+        self.assertEqual(sorted(courses), sorted(expected_course_list))
+
+    def test_successful_rerun(self):
+        """Tests a successful rerun and 2nd rerun"""
+        api.copy_course_videos('test-course', 'meow-rerun')
+        videos = Video.objects.all()
+        for video in videos:
+            self._check_video_course_list(video,
+                                          ['test-course',
+                                           'meow-rerun'])
```
gwprice: Fix whitespace

edxval/tests/test_api.py
```diff
@@ -436,3 +622,74 @@ def test_get_info_queries_for_only_video(self):
...
+
+        api.copy_course_videos('meow-rerun', 'woof-bark')
+        videos = Video.objects.all()
+        for video in videos:
+            self._check_video_course_list(video,
+                                          ['test-course',
+                                           'meow-rerun',
+                                           'woof-bark'])
```
gwprice: Fix whitespace

edxval/tests/constants.py
BenjiLee: Will address this when reorganizing constants/unit tests.

edxval/tests/test_api.py
BenjiLee: Will address this when reorganizing constants/unit tests.

edxval/tests/test_api.py
gwprice: This docstring needs to be updated

edxval/tests/test_api.py
```diff
@@ -436,3 +641,75 @@ def test_get_info_queries_for_only_video(self):
...
+        api.copy_course_videos('test-course', 'course-rerun1')
+        original_videos = Video.objects.filter(courses__course_id='test-course')
+        copied_videos = Video.objects.filter(courses__course_id='course-rerun1')
+        other_course = Video.objects.filter(courses__course_id='test-course2')
+
+        self.assertEqual(len(original_videos), 2)
+        self.assertEqual(original_videos[0].edx_video_id,
+                         constants.VIDEO_DICT_FISH['edx_video_id'])
+        self.assertEqual(original_videos[1].edx_video_id,
+                         constants.VIDEO_DICT_STAR['edx_video_id'])
```
gwprice: The order is not guaranteed. Use set comparison:

```
self.assertEqual(
    {original_video.edx_video_id for original_video in original_videos},
    {constants.VIDEO_DICT_FISH["edx_video_id"], constants.VIDEO_DICT_STAR["edx_video_id"]}
)
```

edxval/tests/test_api.py
```diff
@@ -436,3 +641,75 @@ def test_get_info_queries_for_only_video(self):
         with self.assertNumQueries(4):
             api.get_video_info(constants.VIDEO_DICT_ZEBRA.get("edx_video_id"))
 
+
+class TestCopyCourse(TestCase):
+    """Tests rerun course"""
```
nasthagiri: Consider using consistent terminology, replacing all "rerun"s with "copy"s.

edxval/tests/test_api.py
```diff
@@ -436,3 +641,75 @@ def test_get_info_queries_for_only_video(self):
...
+        self.assertEqual(len(original_videos), 2)
+        self.assertEqual(original_videos[0].edx_video_id,
+                         constants.VIDEO_DICT_FISH['edx_video_id'])
+        self.assertEqual(original_videos[1].edx_video_id,
+                         constants.VIDEO_DICT_STAR['edx_video_id'])
+        self.assertTrue(set(original_videos) == set(copied_videos))
+
+        self.assertEqual(len(other_course), 1)
+        self.assertEqual(other_course[0].edx_video_id,
+                         constants.VIDEO_DICT_TREE['edx_video_id'])
```
gwprice: I don't think this part is necessary; it's enough that the extra video does not appear in the copied_videos list.

edxval/tests/test_api.py
```diff
@@ -436,3 +641,75 @@ def test_get_info_queries_for_only_video(self):
...
+    def test_same_course_ids(self):
+        """Tests when the new course id name is the same as the original"""
+        original_videos = Video.objects.filter(courses__course_id='test-course')
+        api.copy_course_videos('test-course', 'test-course')
+        copied_videos = Video.objects.filter(courses__course_id='test-course')
+        self.assertEqual(len(original_videos), 2)
+        self.assertEqual(original_videos[0].edx_video_id,
+                         constants.VIDEO_DICT_FISH['edx_video_id'])
+        self.assertEqual(original_videos[1].edx_video_id,
+                         constants.VIDEO_DICT_STAR['edx_video_id'])
```
gwprice: I don't think these are really necessary; the only assertion that really matters is the one on the next line.

edxval/api.py
```diff
@@ -296,11 +343,39 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+        return_dict[enc_vid.video.edx_video_id].setdefault("profiles", {}).update(
+            {enc_vid.profile.profile_name: {
+                "url": enc_vid.url,
+                "file_size": enc_vid.file_size,
+            }}
+        )
+    return return_dict
+
+
+def copy_course_videos(old_course_id, new_course_id):
```
gwprice: s/old/source/
s/new/dest/

edxval/tests/test_api.py
```diff
@@ -436,3 +641,75 @@ def test_get_info_queries_for_only_video(self):
...
+        self.assertTrue(set(original_videos) == set(copied_videos))
+
+    def test_existing_new_course_id_for_rerun(self):
```
gwprice: "existing_new_course_id" is kind of confusing, and we shouldn't refer to "rerun" here. I suggest test_existing_destination_course

edxval/tests/test_api.py
```diff
@@ -436,3 +641,75 @@ def test_get_info_queries_for_only_video(self):
...
+        self.assertEqual(original_videos[1].edx_video_id,
+                         constants.VIDEO_DICT_STAR['edx_video_id'])
+        self.assertEqual(len(copied_videos), 3)
+
+        self.assertEqual(copied_videos[0].edx_video_id,
+                         constants.VIDEO_DICT_FISH['edx_video_id'])
+        self.assertEqual(copied_videos[1].edx_video_id,
+                         constants.VIDEO_DICT_STAR['edx_video_id'])
+        self.assertEqual(copied_videos[2].edx_video_id,
+                         constants.VIDEO_DICT_TREE['edx_video_id'])
```
gwprice: Instead of these specific assertions, you can just do self.assertLessEqual(set(original_videos), set(copied_videos)); the <= operator on sets does subset comparison.

gwprice: (I do think the length assertions are good to keep)

edxval/api.py
```diff
@@ -296,11 +343,39 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+                "url": enc_vid.url,
+                "file_size": enc_vid.file_size,
+            }}
+        )
+    return return_dict
+
+
+def copy_course_videos(source_course_id, destination_course_id):
+    """
+    Adds the new_course_id to the videos taken from the old_course_id
```
gwprice: Update

edxval/api.py
```diff
@@ -296,11 +343,39 @@ def get_video_info_for_course_and_profile(course_id, profile_name):
...
+
+    Args:
+        old_course_id: The original course_id
+        new_course_id: The new course_id where the videos will be copied
```
gwprice: Update

edxval/tests/test_api.py
```diff
@@ -436,3 +641,57 @@ def test_get_info_queries_for_only_video(self):
         with self.assertNumQueries(4):
             api.get_video_info(constants.VIDEO_DICT_ZEBRA.get("edx_video_id"))
 
+
+class TestCopyCourse(TestCase):
+    """Tests copy_course_videos in api.py"""
+
+    def setUp(self):
+        """
+        Creates a a course with 2 videos and a course with 1 video
```
gwprice: s/a a/a/

edxval/tests/test_api.py
```diff
@@ -269,6 +270,210 @@ def test_get_url_for_profile(self):
...
+                constants.PROFILE_DICT_DESKTOP["profile_name"]: constants.ENCODED_VIDEO_DICT_DESKTOP2,
+            }
+        ))
+        self.assertEqual(videos, expected_dict)
+
+    def test_get_video_for_course_profiles_no_profile(self):
+        """Tests get_video_info_for_course_and_profiles with no such profile"""
+        videos = api.get_video_info_for_course_and_profiles(
+            'test-course',
+            ['no profile'])
```
nasthagiri: do we allow spaces in profile names?  or did you mean this to be `no_profile`?

edxval/api.py
```diff
@@ -277,17 +278,63 @@ def get_videos_for_ids(
...
+        }
+    Example:
+        Given two videos with two profiles each in course_id 'test_course':
+        {
+            u'edx_video_id_1': {
+                u'duration: 1111,
+                u'profiles': {
+                    u'mobile': {
+                    'url': u'http: //www.example.com/meow',
+                    'file_size': 2222
```
gwprice: Indent

nasthagiri:
:+1: lgtm.  waiting to see cleanup of the test code via consolidation of constants.py in the future.

edxval/tests/test_api.py
BenjiLee: I don't think so. It should be no_profile.

gwprice:
:+1:

PR #42, opened 2015-03-24
Remove Profile fields other than name

gwprice:
None of the fields except profile_name is ever used at present (we don't
even do any validation that the encoded videos added to VAL match the
profile), and the edX media team prefers that the profiles be somewhat
fungible anyway. The motivation for this change is to avoid the need to
sync profile information between systems when doing export/import of
video modules that use Video IDs.

@BenjiLee @ormsbee please review (and please do actually look at the migration. It has been manually edited.)

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -3,11 +3,14 @@
 """
 The internal API for VAL. This is not yet stable
 """
-from enum import Enum
 import logging
 
-from edxval.models import Video, EncodedVideo, CourseVideo
-from edxval.serializers import VideoSerializer, ProfileSerializer
+from enum import Enum
+
+from django.core.exceptions import ValidationError
+
+from edxval.models import Video, EncodedVideo, CourseVideo, Profile
+from edxval.serializers import VideoSerializer
 
 logger = logging.getLogger(__name__)  # pylint: disable=C0103
 
@@ -88,11 +91,7 @@ def create_video(video_data):
                 encoded_video: a list of EncodedVideo dicts
                     url: url of the video
                     file_size: size of the video in bytes
-                    profile: a dict of encoding details
-                        profile_name: ID of the profile
-                        extension: 3 letter extension of video
-                        width: horizontal pixel resolution
-                        height: vertical pixel resolution
+                    profile: ID of the profile
                 courses: Courses associated with this video
          }
     """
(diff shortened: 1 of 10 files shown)


ormsbee:
:+1:

edxval/models.py
```diff
@@ -54,9 +54,6 @@ class Profile(models.Model):
             ),
         ]
     )
-    extension = models.CharField(max_length=10)
-    width = models.PositiveIntegerField()
-    height = models.PositiveIntegerField()
```
BenjiLee: Update the doc at the top of this file.

edxval/migrations/0004_auto__chg_field_profile_width__chg_field_profile_extension__chg_field_.py
```diff
@@ -0,0 +1,74 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # The extension, width, and height fields of the Profile model are
+        # removed, but we want this migration to be non-destructive. Thus, the
+        # columns are simply altered to allow null values.
```
BenjiLee: Any reason why this is not a doc string?

gwprice: No, will fix

edxval/models.py
gwprice: Will do

edxval/api.py
```diff
@@ -88,11 +91,7 @@ def create_video(video_data):
                 encoded_video: a list of EncodedVideo dicts
                     url: url of the video
                     file_size: size of the video in bytes
-                    profile: a dict of encoding details
-                        profile_name: ID of the profile
-                        extension: 3 letter extension of video
-                        width: horizontal pixel resolution
-                        height: vertical pixel resolution
+                    profile: ID of the profile
```
BenjiLee: I think this needs to be `profile_name`

edxval/api.py
```diff
@@ -154,11 +144,7 @@ def get_video_info(edx_video_id, location=None):  # pylint: disable=W0613
...
```
BenjiLee: `profile_name`

BenjiLee:
In the EncodedVideo model, should the field `profile` be changed to `profile_name`?

gwprice:
Changing profile to profile_name would be a backwards incompatible change, and I don't think it's necessary

edxval/migrations/0004_auto__chg_field_profile_width__chg_field_profile_extension__chg_field_.py
gwprice: Actually, this reads strangely as a docstring to me. I don't think it's necessary to include a docstring for a migration function, since they are never really called except by the migration tooling.

BenjiLee:
`from django.core.exceptions import ValidationError` in test_api.py no longer used.

BenjiLee:
Goodbye profiles :+1:

BenjiLee:
Goodby

gwprice:
Updated

gwprice:
I have added custom SQL statements for mysql. Please review.

BenjiLee:
:+1:

PR #44, opened 2015-03-31
Add XML export and import API functions

gwprice:
These will be used to support export/import of courses containing video
modules that rely on VAL being populated. This does not allow updating
a video (including adding encodings) via import.

JIRA: MA-110

I created a new branch (and, therefore, PR) to avoid clobbering Nimisha's work.

@nasthagiri @BenjiLee Please review

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -5,6 +5,7 @@ The internal API for VAL. This is not yet stable
 """
 import logging
 
+from lxml.etree import Element, SubElement
 from enum import Enum
 
 from django.core.exceptions import ValidationError
@@ -123,12 +124,28 @@ def create_profile(profile_name):
         raise ValCannotCreateError(err.message_dict)
 
 
-def get_video_info(edx_video_id, location=None):  # pylint: disable=W0613
+def _get_video(edx_video_id):
+    """
+    Get a Video instance, prefetching encoded video and course information.
+
+    Raises ValVideoNotFoundError if the video cannot be retrieved.
+    """
+    try:
+        return Video.objects.prefetch_related("encoded_videos", "courses").get(edx_video_id=edx_video_id)
+    except Video.DoesNotExist:
+        error_message = u"Video not found for edx_video_id: {0}".format(edx_video_id)
+        raise ValVideoNotFoundError(error_message)
+    except Exception:
+        error_message = u"Could not get edx_video_id: {0}".format(edx_video_id)
+        logger.exception(error_message)
+        raise ValInternalError(error_message)
+
+
+def get_video_info(edx_video_id):
     """
     Retrieves all encoded videos of a video found with given video edx_video_id
 
     Args:
-        location (str): geographic locations used determine CDN
         edx_video_id (str): id for video content.
 
     Returns:
@@ -175,17 +192,7 @@ def get_video_info(edx_video_id, location=None):  # pylint: disable=W0613
             ]
         }
     """
-    try:
-        video = Video.objects.prefetch_related("encoded_videos", "courses").get(edx_video_id=edx_video_id)
-        result = VideoSerializer(video)
-    except Video.DoesNotExist:
-        error_message = u"Video not found for edx_video_id: {0}".format(edx_video_id)
-        raise ValVideoNotFoundError(error_message)
-    except Exception:
-        error_message = u"Could not get edx_video_id: {0}".format(edx_video_id)
-        logger.exception(error_message)
-        raise ValInternalError(error_message)
-    return result.data  # pylint: disable=E1101
+    return VideoSerializer(_get_video(edx_video_id)).data
 
 
 def get_urls_for_profiles(edx_video_id, profiles):
@@ -365,3 +372,96 @@ def copy_course_videos(source_course_id, destination_course_id):
             video=video,
             course_id=destination_course_id
         )
+
+
+def export_to_xml(edx_video_id):
+    """
+    Exports data about the given edx_video_id into the given xml object.
+
+    Args:
+        edx_video_id (str): The ID of the video to export
+
+    Returns:
+        An lxml video_asset element containing export data
+
+    Raises:
+        ValVideoNotFoundError: if the video does not exist
+    """
+    video = _get_video(edx_video_id)
+    video_el = Element(
+        'video_asset',
+        attrib={
+            'client_video_id': video.client_video_id,
+            'duration': unicode(video.duration),
+        }
+    )
+    for encoded_video in video.encoded_videos.all():
+        SubElement(
+            video_el,
+            'encoded_video',
+            {
+                name: unicode(getattr(encoded_video, name))
+                for name in ['profile', 'url', 'file_size', 'bitrate']
+            }
+        )
+    # Note: we are *not* exporting Subtitle data since it is not currently updated by VEDA or used
+    # by LMS/Studio.
+    return video_el
+
+
+def import_from_xml(xml, edx_video_id, course_id=None):
+    """
+    Imports data from a video_asset element about the given edx_video_id.
+
+    If the edx_video_id already exists, then no changes are made. If an unknown
+    profile is referenced by an encoded video, that encoding will be ignored.
+
+    Args:
+        xml: An lxml video_asset element containing import data
+        edx_video_id (str): The ID for the video content
+        course_id (str): The ID of a course to associate the video with
+            (optional)
+
+    Raises:
+        ValCannotCreateError: if there is an error importing the video
+    """
+    if xml.tag != 'video_asset':
+        raise ValCannotCreateError('Invalid XML')
+
+    if Video.objects.filter(edx_video_id=edx_video_id).exists():
+        logger.info(
+            "edx_video_id '%s' present in course '%s' not imported because it exists in VAL.",
+            edx_video_id,
+            course_id,
+        )
+        return
+
+    data = {
+        'edx_video_id': edx_video_id,
+        'client_video_id': xml.get('client_video_id'),
+        'duration': xml.get('duration'),
+        'status': 'imported',
+        'encoded_videos': [],
+        'courses': [],
+    }
+    for encoded_video_el in xml.iterfind('encoded_video'):
+        profile_name = encoded_video_el.get('profile')
+        try:
+            Profile.objects.get(profile_name=profile_name)
+        except Profile.DoesNotExist:
+            logger.info(
+                "Imported edx_video_id '%s' contains unknown profile '%s'.",
+                edx_video_id,
+                profile_name
+            )
+            continue
+        data['encoded_videos'].append({
+            'profile': profile_name,
+            'url': encoded_video_el.get('url'),
+            'file_size': encoded_video_el.get('file_size'),
+            'bitrate': encoded_video_el.get('bitrate'),
+        })
+    if course_id:
+        data['courses'].append(course_id)
+
+    create_video(data)
--- a/edxval/serializers.py
+++ b/edxval/serializers.py
@@ -79,8 +79,9 @@ class CourseSerializer(serializers.RelatedField):
 
     def from_native(self, data):
         if data:
-            return CourseVideo(course_id=data)
-
+            course_video = CourseVideo(course_id=data)
+            course_video.full_clean(exclude=["video"])
+            return course_video
 
 class VideoSerializer(serializers.ModelSerializer):
     """
--- a/edxval/tests/test_api.py
+++ b/edxval/tests/test_api.py
@@ -4,6 +4,7 @@ Tests for the API for Video Abstraction Layer
 """
 
 import mock
+from lxml import etree
 
 from django.test import TestCase
 from django.db import DatabaseError
@@ -16,9 +17,9 @@ from edxval import api as api
 from edxval.api import (
     SortDirection,
     ValCannotCreateError,
+    ValVideoNotFoundError,
     VideoSortField,
 )
-from edxval.serializers import VideoSerializer
 from edxval.tests import constants, APIAuthTestCase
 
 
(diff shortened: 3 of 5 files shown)


edxval/api.py
```diff
@@ -123,6 +124,23 @@ def create_profile(profile_name):
...
+    except Video.DoesNotExist:
+        error_message = u"Video not found for edx_video_id: {0}".format(edx_video_id)
+        raise ValVideoNotFoundError(error_message)
+    except Exception:
+        error_message = u"Could not get edx_video_id: {0}".format(edx_video_id)
+        logger.exception(error_message)
+        raise ValInternalError(error_message)
+
+
 def get_video_info(edx_video_id, location=None):  # pylint: disable=W0613
```
nasthagiri: can we remove the unused parameter?

edxval/api.py
```diff
@@ -365,3 +373,99 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+
+def import_from_xml(xml, edx_video_id, course_id=None):
+    """
+    Imports data from a video_asset element about the given edx_video_id.
+
+    If the edx_video_id already exists, then no changes are made. If an unknown
+    profile is referenced by an encoded video, that encoding will be ignored.
+
+    Args:
+        xml: An lxml video_asset_data element containing import data
```
nasthagiri: s/video_asset_data/video_asset

edxval/api.py
```diff
@@ -365,3 +373,99 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+        )
+        return
+    except ValVideoNotFoundError:
+        pass
+
+    data = {
+        'edx_video_id': edx_video_id,
+        'client_video_id': xml.get('client_video_id'),
+        'duration': xml.get('duration'),
+        'status': 'imported',
```
nasthagiri: why is the `status` value not kept during the import?  The AssetMetaDataStore is imported with the course and would contain the uploaded video information.

edxval/api.py
```diff
@@ -365,3 +373,99 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+        profile_name = encoded_video_el.get('profile')
+        try:
+            profile = Profile.objects.get(profile_name=profile_name)
+        except Profile.DoesNotExist:
+            logger.info(
+                "Imported edx_video_id '%s' contains unknown profile '%s'.",
+                edx_video_id,
+                profile_name
+            )
+            continue
```
nasthagiri: why are we not importing profiles automatically?
by not doing so, it requires manual intervention and careful manual process to make sure profile information is updated before import, etc.  we can be easily flexible about this by automatically creating the profile (and logging as such).

edxval/api.py
```diff
@@ -365,3 +373,99 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+            (optional)
+
+    Raises:
+        ValCannotCreateError: if there is an error importing the video
+    """
+    if xml.tag != 'video_asset':
+        raise ValCannotCreateError('Invalid XML')
+
+    try:
+        video = _get_video(edx_video_id)
```
nasthagiri: If we are just checking for the pre-existence of the video here, then why call the `_get_video` method that prefetches all the related tables?  It's unnecessarily wieldy for an existence check.

edxval/serializers.py
```diff
@@ -79,8 +79,9 @@ def to_native(self, value):
 
     def from_native(self, data):
         if data:
-            return CourseVideo(course_id=data)
-
+            course_video = CourseVideo(course_id=data)
+            course_video.full_clean(exclude=["video"])
```
nasthagiri: Why was this needed?

edxval/tests/test_api.py
```diff
@@ -687,3 +678,240 @@ def test_existing_video_in_destination_course_id(self):
...
+    def assert_xml_equal(self, left, right):
+        """
+        Assert that the given XML fragments have the same attributes, text, and
+        (recursively) children
+        """
+        def get_child_tags(elem):
+            """Extract the list of tag names for children of elem"""
+            return [child.tag for child in elem]
+
+        for attr in ['tag', 'attrib', 'text', 'tail']:
```
nasthagiri: does this verify the other XML attributes: `'profile', 'url', 'file_size', 'nitrate'`?

edxval/tests/test_api.py
```diff
@@ -687,3 +678,240 @@ def test_existing_video_in_destination_course_id(self):
...
+            **constants.ENCODED_VIDEO_DICT_MOBILE
+        )
+        CourseVideo.objects.create(video=video, course_id='existing_course_id')
+
+    def make_import_xml(self, video_dict, encoded_video_dicts=None):
+        ret = etree.Element(
+            "video_asset",
+            attrib={
+                key: unicode(video_dict[key])
+                for key in ["client_video_id", "duration"]
```
nasthagiri: what about `status`?

edxval/api.py
gwprice: Will do

edxval/api.py
gwprice: Good catch. Will fix.

edxval/api.py
gwprice: Good point. I had used this function for the change with additive merging because the encoded videos and courses were needed, but that is no longer the case.

edxval/api.py
gwprice: The status seems particular to a specific deployment. If something is in progress on edx.org, it is not necessarily in progress on another platform. It seems to me like this is a clear way of indicating that the video resulted from an import, and there will be no further updates without something in the target system specifically acting on the video.

edxval/api.py
gwprice: IIRC, after discussion with the media team, we wanted to leave control over the profiles with them and not with course teams importing data. However, this may be an artifact from back when we didn't want to automatically create profiles with silly default values (for width, height, and extension). I will circle back with them for further discussion.

edxval/serializers.py
gwprice: Without this, the course_id is never actually validated. Specifically, ImportTest.test_invalid_course_id fails without this.

edxval/tests/test_api.py
gwprice: `attr` here refers to a Python object attribute. The XML attributes are in `attrib` and are compared as a whole dict.

edxval/api.py
```diff
@@ -365,3 +373,99 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+        logger.info(
+            "edx_video_id '%s' present in course '%s' not imported because it exists in VAL.",
+            edx_video_id,
+            course_id,
```
BenjiLee: Should .format() be used instead?

edxval/tests/test_api.py
nasthagiri: ah.. thanks.

edxval/serializers.py
nasthagiri: good catch.

edxval/api.py
gwprice: A quick survey of our code reveals that we use both for logging, with the majority using this style. I don't have a strong opinion either way.

edxval/tests/test_api.py
```diff
@@ -687,3 +678,240 @@ def test_existing_video_in_destination_course_id(self):
...
+        """)
+        self.assert_xml_equal(
+            api.export_to_xml(constants.VIDEO_DICT_FISH["edx_video_id"]),
+            expected
+        )
+
+    def test_unknown_video(self):
+        with self.assertRaises(ValVideoNotFoundError):
+            api.export_to_xml("unknown_video")
+
```
BenjiLee: newline. Dunno if this travis will fail you for this.

edxval/tests/test_api.py
```diff
@@ -16,6 +17,7 @@
 from edxval.api import (
     SortDirection,
     ValCannotCreateError,
+    ValVideoNotFoundError,
     VideoSortField,
 )
 from edxval.serializers import VideoSerializer
```
BenjiLee: unused import

test-requirements.txt
```diff
@@ -4,3 +4,4 @@ mock==1.0.1
 django-debug-toolbar==1.2.1
 pylint==1.3.0
 ddt==0.8.0
+lxml==3.3.6
```
BenjiLee: Shouldn't this be in `requirements.txt`?

edxval/tests/test_api.py
gwprice: Will fix

edxval/api.py
gwprice: I talked with Greg M, and he said that the profiles should be managed by them and not automatically created by course imports.

test-requirements.txt
gwprice: Yes

edxval/tests/test_api.py
gwprice: Good catch

edxval/api.py
gwprice: We discussed this in person and will leave as is if we can make it work reasonably with the video upload page in Studio.

edxval/tests/test_api.py
```diff
@@ -687,3 +677,241 @@ def test_existing_video_in_destination_course_id(self):
...
+
+    def test_unknown_video(self):
+        with self.assertRaises(ValVideoNotFoundError):
+            api.export_to_xml("unknown_video")
+
+
+class ImportTest(TestCase):
+    """Tests import_from_xml"""
+    def setUp(self):
+        mobile_profile=Profile.objects.create(profile_name=constants.PROFILE_MOBILE)
```
BenjiLee: nitpick: `mobile_profile = Profile...`

edxval/tests/test_api.py
```diff
@@ -687,3 +678,240 @@ def test_existing_video_in_destination_course_id(self):
...
+
+        video = Video.objects.get(edx_video_id=constants.VIDEO_DICT_STAR["edx_video_id"])
+        self.assertFalse(video.encoded_videos.filter(profile__profile_name=profile).exists())
+
+    def test_invalid_tag(self):
+        xml = etree.Element(
+            "invalid_tag",
+            attrib={
+                "client_video_id": "dummy",
+                "duration": "0",
```
BenjiLee: Why is this 0 a string?

gwprice: It's an XML attribute

edxval/tests/test_api.py
gwprice: Will fix

edxval/tests/test_api.py
```diff
@@ -687,3 +677,241 @@ def test_existing_video_in_destination_course_id(self):
...
+        )
+        self.assert_invalid_import(xml)
+
+    def test_invalid_encoded_video_attr(self):
+        xml = self.make_import_xml(
+            video_dict=constants.VIDEO_DICT_FISH,
+            encoded_video_dicts=[{
+                "url": "http://example.com/dummy",
+                "file_size": -1,
+                "bitrate": 0,
```
BenjiLee: Same test for negative bitrate?

gwprice: I don't think that would provide much additional value. If we test each attribute separately, then we are really just doing an exhaustive test of the VideoSerializer, which should be done elsewhere.

gwprice: OTOH, it would be pretty easy to do with ddt

BenjiLee: Then should this test also be put to said other place. (I'm assuming we are testing the negative filesize)

edxval/api.py
```diff
@@ -365,3 +372,96 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+        'client_video_id': xml.get('client_video_id'),
+        'duration': xml.get('duration'),
+        'status': 'imported',
+        'encoded_videos': [],
+        'courses': [],
+    }
+    for encoded_video_el in xml.iterfind('encoded_video'):
+        profile_name = encoded_video_el.get('profile')
+        try:
+            profile = Profile.objects.get(profile_name=profile_name)
```
BenjiLee: unused variable `profile`

edxval/api.py
```diff
@@ -365,3 +372,96 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+    video = _get_video(edx_video_id)
+    video_el = Element(
+        'video_asset',
+        attrib={
+            'client_video_id': video.client_video_id,
+            'duration': unicode(video.duration),
+        }
+    )
+    for encoded_video in video.encoded_videos.all():
+         SubElement(
```
BenjiLee: indent is 5 spaces

edxval/tests/test_api.py
gwprice: Those tests already exist

BenjiLee: Ah I see, we are testing a general invalid xml in this case.

edxval/api.py
gwprice: Will fix

edxval/api.py
gwprice: Will fix

BenjiLee:
Happy so far. I'll take another look tonight.

edxval/tests/test_api.py
```diff
@@ -687,3 +677,241 @@ def test_existing_video_in_destination_course_id(self):
...
+            profile=mobile_profile,
+            **constants.ENCODED_VIDEO_DICT_MOBILE
+        )
+        EncodedVideo.objects.create(
+            video=video,
+            profile=desktop_profile,
+            **constants.ENCODED_VIDEO_DICT_DESKTOP
+        )
+
+    def assert_xml_equal(self, left, right):
```
nasthagiri: we should eventually move this to a common utility library.

edxval/tests/test_api.py
```diff
@@ -687,3 +677,241 @@ def test_existing_video_in_destination_course_id(self):
...
+        video = Video.objects.get(edx_video_id=constants.VIDEO_DICT_FISH["edx_video_id"])
+        self.assert_video_matches_dict(video, constants.VIDEO_DICT_FISH)
+        self.assert_encoded_video_matches_dict(
+            video.encoded_videos.get(profile__profile_name=constants.PROFILE_MOBILE),
+            constants.ENCODED_VIDEO_DICT_MOBILE
+        )
+        self.assertFalse(
+            video.encoded_videos.filter(profile__profile_name=constants.PROFILE_DESKTOP).exists()
+        )
+        self.assertFalse(video.courses.filter(course_id=new_course_id).exists())
```
nasthagiri: If the video already existed, for performance reasons, the `course_id` should have still been added to the list of courses.

nasthagiri:
:+1: except for: "If the video already existed, for performance reasons, the course_id should have still been added to the list of courses."

edxval/tests/test_api.py
gwprice: That would be nice but is made more complicated by the fact that the CourseSerializer is not in fact a serializer but a field, so I punted on that.

BenjiLee:
obligatory :+1:

PR #45, opened 2015-04-01
MA-410 Add course_id on import for existing VAL entry.

nasthagiri:
https://openedx.atlassian.net/browse/MA-410

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -428,21 +428,31 @@ def import_from_xml(xml, edx_video_id, course_id=None):
     if xml.tag != 'video_asset':
         raise ValCannotCreateError('Invalid XML')
 
-    if Video.objects.filter(edx_video_id=edx_video_id).exists():
+    # If video with edx_video_id already exists, associate it with the given course_id.
+    try:
+        video = Video.objects.get(edx_video_id=edx_video_id)
         logger.info(
             "edx_video_id '%s' present in course '%s' not imported because it exists in VAL.",
             edx_video_id,
             course_id,
         )
+        if course_id:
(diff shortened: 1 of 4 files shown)


nasthagiri:
@gwprice @BenjiLee Please review.

edxval/api.py
```diff
@@ -428,21 +428,30 @@ def import_from_xml(xml, edx_video_id, course_id=None):
...
+        video = _get_video(edx_video_id)
         logger.info(
             "edx_video_id '%s' present in course '%s' not imported because it exists in VAL.",
             edx_video_id,
             course_id,
         )
+        if course_id:
+            CourseVideo.objects.get_or_create(
+                video=video,
+                course_id=course_id
```
gwprice: course_id needs to be validated

edxval/api.py
```diff
@@ -428,21 +428,30 @@ def import_from_xml(xml, edx_video_id, course_id=None):
     if xml.tag != 'video_asset':
         raise ValCannotCreateError('Invalid XML')
 
-    if Video.objects.filter(edx_video_id=edx_video_id).exists():
+    try:
+        video = _get_video(edx_video_id)
```
gwprice: _get_video does unnecessary prefetching. It might be best to add a prefetch argument to _get_video.

BenjiLee:
:+1: when Greg's comments are addressed.

nasthagiri:
@gwprice Please re-review.

gwprice:
:+1:

gwprice:
I just realized that there should be another test case for this change (one that I think will currently fail). A video that is already associated with a specific course should be re-imported for that course.

PR #48, opened 2015-08-28
Update Django rest framework

wedaly:
* Upgrade Django Rest Framework to version 3.1
* Add an external library for OAuth authentication (moved out of DRF core)
* Add a test case for updating course IDs associated with a video.

Part of the Django 1.8 upgrade work.  This introduces one change to the API: previously, when PUTing a video, `courses` and `encoded_videos` not specified in the data would not be deleted.  This may or may not be a bug in the current implementation -- if it's not, I'll need to update the test case I added and fix the code.

### DIFF
--- a/edxval/serializers.py
+++ b/edxval/serializers.py
@@ -5,7 +5,7 @@ Serialization is usually sent through the VideoSerializer which uses the
 EncodedVideoSerializer which uses the profile_name as it's profile field.
 """
 from rest_framework import serializers
-from django.core.exceptions import ValidationError
+from rest_framework.fields import IntegerField, DateTimeField
 
 from edxval.models import Profile, Video, EncodedVideo, Subtitle, CourseVideo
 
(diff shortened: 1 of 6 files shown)


edxval/tests/test_views.py
```diff
@@ -278,7 +278,7 @@ def test_update_courses(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK, msg=response.content)
 
         # Verify that the video's courses were updated
-        # Note: the current version of edx-val does NOT remove old course videos!
+        # Note: the current version of edx-val does NOT remove old course videosal!
```
BenjiLee: typo in `videosa1`?

wedaly: That's the problem with using `vim`...  Updating this now.

edxval/serializers.py
```diff
@@ -44,20 +57,17 @@ class SubtitleSerializer(serializers.ModelSerializer):
...
     content = serializers.CharField(write_only=True)
 
-    def validate_content(self, attrs, source):
-        """
-        Validate that the subtitle is in the correct format
-        """
-        value = attrs[source]
-        if attrs.get('fmt') == 'sjson':
+    def validate(self, data):
+        value = data.get("content")
```
BenjiLee: nit: Docstring

wedaly: Updated.

edxval/tests/test_serializers.py
```diff
@@ -137,7 +145,8 @@ def test_wrong_input_type(self):
         """
         data = "hello"
         serializer = VideoSerializer(data=data)
+        self.assertFalse(serializer.is_valid())
         self.assertEqual(
             serializer.errors.get("non_field_errors")[0],
-            "Invalid data"
+            "Invalid data. Expected a dictionary, but got str."
```
BenjiLee: Is the change in the error message a DRF3 change?

wedaly: Yes.

BenjiLee:
LGTM :+1:

PR #50, opened 2015-09-09
Add regression tests in preparation for the Django Rest Framework upgrade

wedaly:
@BenjiLee could you please review this when you have a moment.  Just adding some test cases in preparation for the django rest framework upgrade.

### DIFF
--- a/edxval/tests/constants.py
+++ b/edxval/tests/constants.py
@@ -268,6 +268,22 @@ COMPLETE_SET_UPDATE_STAR = dict(
     subtitles=[SUBTITLE_DICT_SRT],
     **VIDEO_DICT_STAR
 )
+COMPLETE_SET_WITH_COURSE_KEY = dict(
+    courses=['edX/DemoX/Demo_Course'],
+    encoded_videos=[
+        ENCODED_VIDEO_DICT_STAR
+    ],
+    subtitles=[SUBTITLE_DICT_SRT],
+    **VIDEO_DICT_STAR
+)
+COMPLETE_SET_WITH_OTHER_COURSE_KEYS = dict(
+    courses=['edX/DemoX/Astonomy', 'edX/DemoX/Zoology'],
+    encoded_videos=[
+        ENCODED_VIDEO_DICT_STAR
+    ],
+    subtitles=[SUBTITLE_DICT_SRT],
+    **VIDEO_DICT_STAR
+)
 COMPLETE_SET_NOT_A_LIST = dict(
     encoded_videos=dict(
         url="https://www.howIwonder.com",
--- a/edxval/tests/test_views.py
+++ b/edxval/tests/test_views.py
@@ -6,7 +6,7 @@ from django.core.urlresolvers import reverse
 from rest_framework import status
 
 from edxval.tests import constants, APIAuthTestCase
-from edxval.models import Profile, Video
+from edxval.models import Profile, Video, CourseVideo
 
 
 class VideoDetail(APIAuthTestCase):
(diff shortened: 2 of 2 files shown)


edxval/tests/test_views.py
```diff
@@ -207,6 +207,92 @@ def test_update_one_of_two_encoded_videos(self):
...
+        # Expect that encoded videos have been removed
+        videos = Video.objects.all()
+        self.assertEqual(len(videos), 1)
+        self.assertEqual(len(videos[0].encoded_videos.all()), 0)
+
+    def test_update_courses(self):
+        # Create the video with associated course keys
+        url = reverse('video-list')
+        response = self.client.post(url, constants.COMPLETE_SET_WITH_COURSE_KEY, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
BenjiLee: Could these 3 lines of code be a helper method since it's used multiple times?

wedaly:
@BenjiLee Good call; updated.

edxval/tests/test_views.py
```diff
@@ -207,6 +207,86 @@ def test_update_one_of_two_encoded_videos(self):
         )
         self.assertEqual(len(videos[0].encoded_videos.all()), 1)
 
+    def test_update_remove_subtitles(self):
+        # Create some subtitles
+        self._create_videos(constants.COMPLETE_SET_STAR)
+
+        # Sanity check that the subtitles
```
BenjiLee: Incomplete sentence

edxval/tests/constants.py
```diff
@@ -268,6 +268,22 @@
...
+    encoded_videos=[
+        ENCODED_VIDEO_UPDATE_DICT_STAR
+    ],
+    subtitles=[SUBTITLE_DICT_SRT],
+    **VIDEO_DICT_STAR
+)
+COMPLETE_SET_WITH_OTHER_COURSE_KEYS = dict(
+    courses=['edX/DemoX/Astonomy', 'edX/DemoX/Zoology'],
+    encoded_videos=[
+        ENCODED_VIDEO_UPDATE_DICT_STAR
```
BenjiLee: Might be better to use "ENCODED_VIDEO_DICT_STAR" since UPDATE was originally meant to be used when updating an object.

edxval/tests/constants.py
```diff
@@ -268,6 +268,22 @@
     subtitles=[SUBTITLE_DICT_SRT],
     **VIDEO_DICT_STAR
 )
+COMPLETE_SET_WITH_COURSE_KEY = dict(
+    courses=['edX/DemoX/Demo_Course'],
+    encoded_videos=[
+        ENCODED_VIDEO_UPDATE_DICT_STAR
```
BenjiLee: Might be better to use "ENCODED_VIDEO_DICT_STAR" since UPDATE was originally meant to be used when updating an object.

edxval/tests/test_views.py
```diff
@@ -207,6 +207,86 @@ def test_update_one_of_two_encoded_videos(self):
...
+        videos = Video.objects.all()
+        self.assertEqual(len(videos), 1)
+        self.assertEqual(len(videos[0].encoded_videos.all()), 0)
+
+    def test_update_courses(self):
+        # Create the video with associated course keys
+        self._create_videos(constants.COMPLETE_SET_WITH_COURSE_KEY)
+
+        # Verify that the video was associated with the courses
+        video = Video.objects.get()
```
BenjiLee: Why use .get() instead of .all* like the other tests.

edxval/tests/test_views.py
```diff
@@ -207,6 +207,86 @@ def test_update_one_of_two_encoded_videos(self):
...
+        response = self.client.put(
+            url,
+            constants.COMPLETE_SET_WITH_OTHER_COURSE_KEYS,
+            format='json'
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK, msg=response.content)
+
+        # Verify that the video's courses were updated
+        # Note: the current version of edx-val does NOT remove old course videos!
+        course_keys = [c.course_id for c in CourseVideo.objects.filter(video=video)]
```
BenjiLee: Isn't course_keys already set?

BenjiLee:
Done with first pass.

edxval/tests/test_views.py
wedaly: In this case, I know that there is exactly one video because it was created by the test code (rather than by the code under test).

edxval/tests/test_views.py
wedaly: I'm updating it here with the new values and re-using the variable.

wedaly:
@BenjiLee updated and ready for another pass.

BenjiLee:
:+1:

PR #51, opened 2015-09-11
upgrade to django 1.8

muhammad-ammar:
TNL-2824


muhammad-ammar:
@wedaly any idea why the build is not triggered?

wedaly:
Travis might be down or overloaded?

muhammad-ammar:
@BenjiLee Any specific reason for adding `django-debug-toolbar` in [test-requirements.txt](https://github.com/edx/edx-val/blob/master/test-requirements.txt#L4) and [installed apps](https://github.com/edx/edx-val/blob/master/edxval/settings.py#L134). I am not sure if we are really using it.

muhammad-ammar:
@wedaly @symbolist Please have a look. Travis build is still not running. Locally all the tests are passing.

@macdiesel @muzaffaryousaf FYI

edxval/migrations/0004_remove_profile_fields.py
```diff
@@ -1,89 +0,0 @@
...
-        db.alter_column('edxval_profile', 'height', self.gf('django.db.models.fields.PositiveIntegerField')(null=True, default=1))
-
-        # On MySQL, attempts to insert new model instances (which lack values
-        # for the removed fields) results in a warning about a lack of a default
-        # value. Because django does not have a mechanism for specifying the
-        # default value in the database, we must manually alter the table here.
-        if db.backend_name == 'mysql':
-            db.execute("ALTER TABLE edxval_profile MODIFY extension VARCHAR(10) DEFAULT 'mp4'")
-            db.execute("ALTER TABLE edxval_profile MODIFY width INT(10) UNSIGNED DEFAULT 1")
-            db.execute("ALTER TABLE edxval_profile MODIFY height INT(10) UNSIGNED DEFAULT 1")
```
muhammad-ammar: @wedaly @symbolist Do we need to add these changes in the new migration? Now we will have a single migration and there is no record of these three fields. Any thoughts?

edxval/migrations/0004_remove_profile_fields.py
```diff
@@ -1,89 +0,0 @@
...
-        # Because the forward migration is non-destructive and simply alters the
-        # extension, width, and height columns of the Profile model to allow
-        # null values, values from before the forward migration will still be
-        # present in the table. The backward migration restores the non-null
-        # restriction to each column and provides dumb defaults for any profiles
-        # that were created between the application of the forward migration and
-        # backward migration.
-        db.alter_column('edxval_profile', 'width', self.gf('django.db.models.fields.PositiveIntegerField')(default=1))
-        db.alter_column('edxval_profile', 'extension', self.gf('django.db.models.fields.CharField')(default='mp4', max_length=10))
-        db.alter_column('edxval_profile', 'height', self.gf('django.db.models.fields.PositiveIntegerField')(default=1))
```
muhammad-ammar: @wedaly @symbolist same comment as above.

edxval/migrations/0004_remove_profile_fields.py
muhammad-ammar: @nasthagiri We are upgrading to django 1.8 and south is removed from 1.8 so we need to [upgrade from south](https://docs.djangoproject.com/en/1.8/topics/migrations/#upgrading-from-south). As a result of upgrade a new single migration will be created. 

Before upgrade, initially for the Profile model there were 4 fields but 3 of them get removed and some manual changes(please see above) were introduced for removed fields. Now my question is do we need to add those manual changes to new migration?  As per my understanding we don't need those manual changes because our new migration file don't have any record for those removed fields.

edxval/migrations/0004_remove_profile_fields.py
muhammad-ammar: @nasthagiri same comment as above.

BenjiLee:
@muhammad-ammar No specific reason. Feel free to remove.

muhammad-ammar:
@BenjiLee Thanks!

symbolist:
@muhammad-ammar Do the forward and reverse migrations work correctly? Just want to confirm that this has been tested. Otherwise :+1:.

muhammad-ammar:
@symbolist Not yet. Should we test this on sandbox linked with production? What do you suggest?

symbolist:
I don't think you can run migrations on the prod-clone. Just testing these locally should be adequate.

benpatterson:
Added a commit to get this running on travis (it was not running b/c it was a PR against a shared branch, as opposed to master). I see the build failed, but will leave it to you guys to troubleshoot.

edxval/migrations/0004_remove_profile_fields.py
nasthagiri: @muhammad-ammar So are we requiring all openedx instances to first update to the last edx-platform build BEFORE django 1.8 support, and THEN ONLY update to the next?  If so, then what you say is fine.
However, if we're not doing that, then how will older versions of the platform update their data models correctly?
And if we are doing that, then can we remove all older migration scripts?

edxval/migrations/0004_remove_profile_fields.py
nasthagiri: @muhammad-ammar Same concerns/questions as above.

edxval/migrations/0004_remove_profile_fields.py
nedbat: Our current thinking is that we will need to provide a SQL script to update the database schema from Cypress to Dogwood.

.travis.yml
```diff
@@ -10,5 +10,6 @@ script:
 branches:
     only:
       - master
+      - will/upgrade-django-rest-framework
```
nasthagiri: is this meant to be checked in?

muhammad-ammar: no. currently tests on travis run against master branch so this is needed to run the build.

nasthagiri:
This lgtm given our migration story for open-edx instances.

edxval/migrations/0004_remove_profile_fields.py
macdiesel: @nedbat @nasthagiri I have backlogged a ticket to cover this work: https://openedx.atlassian.net/browse/TNL-3292

nasthagiri: Thanks @macdiesel

muhammad-ammar:
closing in favor of https://github.com/edx/edx-val/pull/53

PR #56, opened 2015-11-09
Convert the data migration to a Django migration

nedbat:


nedbat:
@macdiesel @doctoryes @symbolist @muzaffaryousaf look ok?

edxval/migrations/0002_data__default_profiles.py
```diff
@@ -0,0 +1,36 @@
...
+def create_default_profiles(apps, schema_editor):
+    Profile = apps.get_model("edxval", "Profile")
+    for kwargs in DEFAULT_PROFILES:
+        Profile.objects.get_or_create(**kwargs)
+
+
+def delete_default_profiles(apps, schema_editor):
+    Profile = apps.get_model("edxval", "Profile")
+    for kwargs in DEFAULT_PROFILES:
+        Profile.objects.get_or_create(**kwargs).delete()
```
doctoryes: If you `create()` and `delete()` in the same filter, does it do both in the DB? Or is it smart enough to do nothing?

I think it's fine either way - data migration is a one-time, out-of-band thing.

doctoryes:
:+1:

edxval/migrations/0002_data__default_profiles.py
nedbat: Oh, oops: this won't work: get_or_create returns a pair.... I'll fix it.

nedbat:
@doctoryes thanks for the skepticism, there were a bunch of things wrong with this migration!
Look again?

edxval/migrations/0002_data__default_profiles.py
```diff
@@ -0,0 +1,37 @@
...
+def create_default_profiles(apps, schema_editor):
+    Profile = apps.get_model("edxval", "Profile")
+    for profile in DEFAULT_PROFILES:
+        Profile.objects.get_or_create(profile_name=profile)
+
+
+def delete_default_profiles(apps, schema_editor):
+    Profile = apps.get_model("edxval", "Profile")
+    for profile in DEFAULT_PROFILES:
+        obj, _ = Profile.objects.get_or_create(profile_name=profile)
```
muzaffaryousaf: Why get_or_create ? We need to delete them all. `Profile.objects.filter(profile_name=profile).delete()`

muzaffaryousaf:
@nedbat We already have this data migration in it's initial file. Please see here https://github.com/edx/edx-val/blob/ammar/upgrade-django-1.8/edxval/migrations/0001_initial.py#L9

muzaffaryousaf:
@nedbat We have to remove the data migration from initial file.

muzaffaryousaf:
:+1:

doctoryes:
:+1:

PR #58, opened 2016-02-02
Jibsheet/dont use using

jibsheet:
Having a using() in a migration prevents causes errors when running edx-platform with two backing databases (it forces the migration to try and create/read tables in the wrong database).  Removing the using allows Django to use the internal DB router to find the right location.

I reversed and re-ran this migration in a devstack without issue and checked that since we're updating an already-run migration we won't be impacting future deploys.

I'm also bumping the version so we can update what platform includes in my platform PR.

@BenjiLee 
@edx/devops


feanil:
:+1:

BenjiLee:
:+1:

PR #60, opened 2016-08-23
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -4,4 +4,6 @@
 nick: val
 oeps: {}
 owner: edx/teaching-and-learning
+tags:
+  - video pipline
 track-pulls: true
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,7 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: val
+oeps: {}
+owner: MUST FILL IN OWNER
+track-pulls: true
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -3,5 +3,5 @@
 
 nick: val
 oeps: {}
-owner: MUST FILL IN OWNER
+owner: edx/teaching-and-learning
 track-pulls: true


cpennington:
@edx/teaching-and-learning: I've tagged your team as the owner of this repo. Ok to merge?

cahrens:
I have no idea what this repo is. @ormsbee, can you explain what this repo is used for?

BenjiLee:
When mobile/desktop wants a video, they request all the video encodings for a particular edx_video_id. edx-val is the video abstraction layer (VAL) so that the requestor does not need to specify what encoding they need, but only need to specify video by content.

cahrens:
I guess that is T&L, correct @nasthagiri?

openedx.yaml
```diff
@@ -0,0 +1,9 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: val
+oeps: {}
+owner: edx/teaching-and-learning
```
BenjiLee: @cpennington Can there only be one?

cpennington: Yes, `owner` is a singular value. Should this be @edx/edx-mobile-push instead?

BenjiLee: Thinking about this practically, it would be better as edx/teaching-and-learning. I don't think mobile will be touching this even if we did 'own' this.

nasthagiri: TNL should be the owner of this.

robrap: When we discussed OEP-2 in one of the architecture meetings, I brought up my concern around ownership of edx-platform.  I was told that we could have multiple owners in rare cases, where edx-platform was an obvious example.  

I see the benefit of a single owner who might just have to serve as a point on questions/decisions regarding a repo.  At the same time, it may force us (for better or for worse in certain cases), to use repos as the boundary for components so that ownership could both be well defined and can change owners.

PR #62, opened 2016-10-19
Update video through api

jagonzalr:
### Description

[OSPR-1517](https://openedx.atlassian.net/browse/OSPR-1517)

Update video through api.
### Reviewers

If you've been tagged for review, please check your corresponding box once you've given the 👍
- [ ] Code review: @ormsbee 
- [ ] Code review: @BenjiLee

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -57,6 +57,13 @@ class ValCannotCreateError(ValError):
     pass
 
 
+class ValCannotUpdateError(ValError):
+    """
+    This error is raised when an object cannot be updated
+    """
+    pass
+
+
 class VideoSortField(Enum):
     """An enum representing sortable fields in the Video model"""
     created = "created"
@@ -109,7 +116,8 @@ def update_video(video_data):
     """
     Called on to update Video objects in the database
 
-    update_video is used to update Video objects.
+    update_video is used to update Video objects by the given edx_video_id in the video_data.
+
     Args:
         video_data (dict):
          {
@@ -123,15 +131,24 @@ def update_video(video_data):
                     profile: ID of the profile
                 courses: Courses associated with this video
          }
+
+    Raises:
+        Raises ValVideoNotFoundError if the video cannot be retrieved.
+        Raises ValCannotUpdateError if the video cannot be updated.
     """
 
-    video = _get_video(video_data.get("edx_video_id"))
+    try:
+        video = _get_video(video_data.get("edx_video_id"))
+    except Video.DoesNotExist:
+        error_message = u"Video not found when trying to update video with edx_video_id: {0}".format(video_data.get("edx_video_id")) 
+        raise ValVideoNotFoundError(error_message)
+        
     serializer = VideoSerializer(video, data=video_data)
     if serializer.is_valid():
         serializer.save()
         return video_data.get("edx_video_id")
     else:
-        raise ValCannotCreateError(serializer.errors)
+        raise ValCannotUpdateError(serializer.errors)
 
 
 def create_profile(profile_name):
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -96,7 +96,13 @@ def create_video(video_data):
                 courses: Courses associated with this video
          }
     """
-    serializer = VideoSerializer(data=video_data)
+
+    try:
+        video = _get_video(video_data.get("edx_video_id"))
+        serializer = VideoSerializer(video, data=video_data)
+    except Exception:
+        serializer = VideoSerializer(data=video_data)
+
     if serializer.is_valid():
         serializer.save()
         return video_data.get("edx_video_id")
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -83,7 +83,7 @@ def create_video(video_data):
     creation will be rejected. If the profile is not found in the database, the
     video will not be created.
     Args:
-        data (dict):
+        video_data (dict):
          {
                 url: api url to the video
                 edx_video_id: ID of the video
@@ -111,8 +111,7 @@ def update_video(video_data):
 
     update_video is used to update Video objects.
     Args:
-        edx_video_id (String)
-        data (dict):
+        video_data (dict):
          {
                 url: api url to the video
                 edx_video_id: ID of the video
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -97,12 +97,37 @@ def create_video(video_data):
          }
     """
 
-    try:
-        video = _get_video(video_data.get("edx_video_id"))
-        serializer = VideoSerializer(video, data=video_data)
-    except Exception:
-        serializer = VideoSerializer(data=video_data)
+    serializer = VideoSerializer(data=video_data)
+    if serializer.is_valid():
+        serializer.save()
+        return video_data.get("edx_video_id")
+    else:
+        raise ValCannotCreateError(serializer.errors)
 
+
+def update_video(video_data):
+    """
+    Called on to update Video objects in the database
+
+    update_video is used to update Video objects.
+    Args:
+        edx_video_id (String)
+        data (dict):
+         {
+                url: api url to the video
+                edx_video_id: ID of the video
+                duration: Length of video in seconds
+                client_video_id: client ID of video
+                encoded_video: a list of EncodedVideo dicts
+                    url: url of the video
+                    file_size: size of the video in bytes
+                    profile: ID of the profile
+                courses: Courses associated with this video
+         }
+    """
+
+    video = _get_video(video_data.get("edx_video_id"))
+    serializer = VideoSerializer(video, data=video_data)
     if serializer.is_valid():
         serializer.save()
         return video_data.get("edx_video_id")
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -102,6 +102,11 @@ def create_video(video_data):
                     profile: ID of the profile
                 courses: Courses associated with this video
          }
+
+    Raises:
+        Raises ValCannotCreateError if the video cannot be created.
+
+    Returns the successfully created Video object
     """
 
     serializer = VideoSerializer(data=video_data)
@@ -135,6 +140,8 @@ def update_video(video_data):
     Raises:
         Raises ValVideoNotFoundError if the video cannot be retrieved.
         Raises ValCannotUpdateError if the video cannot be updated.
+
+    Returns the successfully updated Video object
     """
 
     try:
@@ -142,7 +149,7 @@ def update_video(video_data):
     except Video.DoesNotExist:
         error_message = u"Video not found when trying to update video with edx_video_id: {0}".format(video_data.get("edx_video_id")) 
         raise ValVideoNotFoundError(error_message)
-        
+
     serializer = VideoSerializer(video, data=video_data)
     if serializer.is_valid():
         serializer.save()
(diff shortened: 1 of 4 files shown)


edx-webhook:
Thanks for the pull request, @jagonzalr! I've created [OSPR-1517](https://openedx.atlassian.net/browse/OSPR-1517) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).



If you like, you can add yourself to the [AUTHORS](https://github.com/edx/edx-val/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

BenjiLee:
Updating a video should not be done through the create video endpoint. Could we get more context on what you needed?

jagonzalr:
Hi @BenjiLee, we are trying to fix the duration stuff of a video added through Studio. What we have noticed it that the create video endpoint is never call in Studio (or I still don't get it how to call it when saving a video) and all that xml stuff that is in there. So what I'm doing is calling the create video endpoint clicks on the save button of the editor and the function ```editor_saved``` is call. So inside that function I call the ```create_video``` and then in my PR it checks if it already exists or not and creates or updates it. But I could make a different endpoint just for update. I did it that way because apparently there is no way to update the video so if for example you change the url of the video the information will not be updated. And if you delete that video component and add another one then it is not deleted on edxval.

yro:
@jagonzalr Thanks for your PR! 
I'm a little confused as to where you're fixing the duration, and where it's displaying. Are you using VAL as a method to call the mobile app? Can you discuss the use case a little?

BenjiLee:
@jagonzalr Thank you for the context. It looks like you want to fill in the missing metadata when you retrieve videos such as on the mobile apps. I'm not sure if edx-val is the right place for this and might require more discovery to how we would want to approach this.

Let me know if we are on the same page.

Problem: When using the mobile app, when accessing a video, a 0:00 duration is displayed. 

This is the method that collects the video data for the course blocks api where the mobile app gets its video data.
https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/video_module/video_module.py#L880

When a video cannot be found in VAL, it defaults and sets a 0:00 duration https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/video_module/video_module.py#L948

Goal: Have some sort of way to manually add the duration.

jagonzalr:
@BenjiLee we are not using anything at the moment. We saw a thread in Slack that mention that the mobile apps get the duration from edx-val. So when looking at the code we notice that when saving a video on Studio it doesn't do anythnig on edx-val so videos are always 0:00, always. So I change the edx-val to create or update a video because on our code inside Studio I change the xblock of video so it can actually store the duration. Right now it suppose to do it through an xml but I didn't really get it so I just ignore all of it.

nasthagiri:
@jagonzalr The process we use at edx.org is as follows:

1. Course team uploads Video through Studio.
2. Studio uploads the video in an S3 bucket on AWS.
3. Studio creates a barebones shell entry for the video in VAL.
4. A system called VEDA downloads the video from S3 and encodes it to mobile-friendly encodings.  This encoding process computes the video's duration value.
5. VEDA calls VAL to update the video's information, including its duration value, locations to its encodings, etc.  It does so by calling [this REST API](https://github.com/edx/edx-val/blob/master/edxval/views.py#L55).

Now, we intentionally didn't persist any duration values in Studio itself since searching for a duration of 0 allows us to easily find videos that didn't complete the video encoding process.  Otherwise, mobile users have an undesirable performance experience with trying to watch/download large videos that haven't been encoded for mobile.

Now, for your case, since you (currently) do not have access to VEDA, you can choose to call VAL directly to update the data.  However, for your mobile users, you should make sure the videos are expectedly encoded for them.  For some open edx instances, I believe, they chose to not worry about the encoding process and only entered links to low resolution videos that were mobile-friendly into the video xBlock.

jagonzalr:
Ha @nasthagiri that makes sense. So what, when, how with VEDA? Sounds cool.

nasthagiri:
@yro is the VEDA engineer.  We still need to figure out the deployment and operational details to be able to productize and open source VEDA.

yro:
Thanks @nasthagiri .
@jagonzalr If you are trying to encode videos explicitly for mobile, any mp4 or webm encode works, (obviously the smaller size the better).
You can simply send the endpoint url of the mobile-encoded video and the duration to VAL and it will report the duration correctly in the mobile app. (Step 5 above). There's some tricks around setting up the data in the correct format, but you're already in the code, so that should not be too much of a problem for you.
That's the quickest way to get to what you're trying to accomplish.

jagonzalr:
@nasthagiri @yro from what I'm reading you are saying that I should use the REST API instead of the updated version of the api.py file? I think my change is good since you don't have to call an url inside Studio and just call a method since VAL is already installed in the django apps.

yro:
@jagonzalr Fair point, sounds good. Can you add a test, and I'll approve your changes.

yro:
^^ I see that you have a test case, I think it just isn't getting called from the tests, correct me if I'm wrong.

jagonzalr:
@yro you are correct and I delete it because it was causing trouble when running the tests. The problem was that the create video test wasn't call but the update was and I couldn't figure out how to fix that.

edxval/tests/test_api.py
```diff
@@ -117,6 +118,47 @@ def test_create_invalid_video(self, data): # pylint: disable=W0621
...
+        video_data = dict(
+            encoded_videos=[
+                constants.ENCODED_VIDEO_DICT_FISH_MOBILE
+            ],
+            **constants.VIDEO_DICT_FISH_UPDATE
+        )
+        result = api.update_video(video_data)
+        videos = Video.objects.all()
+        self.assertEqual(len(videos), 1)
+        self.assertEqual("super-soaker", result)
```
BenjiLee: This string will pass even if update was not called. I would renamed the name of the updated video.

jagonzalr: @BenjiLee done

edxval/api.py
```diff
@@ -104,6 +105,36 @@ def create_video(video_data):
         raise ValCannotCreateError(serializer.errors)
 
 
+def update_video(video_data):
+    """
+    Called on to update Video objects in the database
+
+    update_video is used to update Video objects.
+    Args:
+        edx_video_id (String)
```
BenjiLee: args doc does not match up.

edxval/api.py
```diff
@@ -104,6 +105,36 @@ def create_video(video_data):
...
+        data (dict):
```
BenjiLee: should be `video_data`. Would you mind updating the doc in  `create_video` above it :)

edxval/tests/test_api.py
jagonzalr: @BenjiLee done

edxval/api.py
jagonzalr: @BenjiLee done

edxval/api.py
jagonzalr: @BenjiLee done

yro:
👍  I'm good with this.

edxval/api.py
```diff
@@ -104,6 +105,35 @@ def create_video(video_data):
...
+                client_video_id: client ID of video
+                encoded_video: a list of EncodedVideo dicts
+                    url: url of the video
+                    file_size: size of the video in bytes
+                    profile: ID of the profile
+                courses: Courses associated with this video
+         }
+    """
+
+    video = _get_video(video_data.get("edx_video_id"))
```
BenjiLee: If we cannot get a video, I think a ValVideoNotFoundError should be raised. This method should not be used for creating a video.

edxval/api.py
```diff
@@ -104,6 +105,35 @@ def create_video(video_data):
         raise ValCannotCreateError(serializer.errors)
 
 
+def update_video(video_data):
+    """
+    Called on to update Video objects in the database
+
+    update_video is used to update Video objects.
```
BenjiLee: Could be clearer in the doc. 

 `update_video is used to update Video objects by the given edx_video_id.`

I think we may want to add some Raise docs depending on the comments below.

edxval/api.py
```diff
@@ -104,6 +105,35 @@ def create_video(video_data):
...
+         }
+    """
+
+    video = _get_video(video_data.get("edx_video_id"))
+    serializer = VideoSerializer(video, data=video_data)
+    if serializer.is_valid():
+        serializer.save()
+        return video_data.get("edx_video_id")
+    else:
+        raise ValCannotCreateError(serializer.errors)
```
BenjiLee: We should make a new error for failed updates.

BenjiLee: Additionally, a test should be written for this case.

edxval/api.py
BenjiLee: `Raises ValVideoNotFoundError if the video cannot be retrieved.`

edxval/api.py
BenjiLee: We need a test case for when the video is not found.

BenjiLee: We could wrap this around a try catch to be more explicit about the error here.
```
try
    video = _get_video(video_data.get("edx_video_id"))
except Video.DoesNotExist:
        error_message = u"Video not found for edx_video_id: {0}".format(edx_video_id)
        raise ValVideoNotFoundError(error_message)
```
Add to doc `Raises ValVideoNotFoundError if the video cannot be retrieved.`

jagonzalr: @BenjiLee ```_get_video``` already returns ```ValVideoNotFoundError``` so I would say keep it DRY. But can update doc that it returns ```ValVideoNotFoundError``` from ```_get_video```

jagonzalr: @BenjiLee done

edxval/api.py
jagonzalr: @BenjiLee done

edxval/api.py
jagonzalr: @BenjiLee already made the try catch 😉

edxval/api.py
```diff
@@ -104,6 +112,45 @@ def create_video(video_data):
...
+                edx_video_id: ID of the video
+                duration: Length of video in seconds
+                client_video_id: client ID of video
+                encoded_video: a list of EncodedVideo dicts
+                    url: url of the video
+                    file_size: size of the video in bytes
+                    profile: ID of the profile
+                courses: Courses associated with this video
+         }
+
```
BenjiLee: Last nit!

Could you add a return doc? 

`Returns the successfully updated Video object`

and then if you could also add that for the create_video doc as well :)

`Returns the successfully created Video object`

BenjiLee:
Awesome. Squashing and merging when tests are complete.

edxval/api.py
jagonzalr: @BenjiLee done

BenjiLee:
I think we need to do a release in order to have this changes applied to your edx platform instance.

jagonzalr:
Probably, also I have my own fork so in the meantime I can point my repo for my instance until the release.

PR #64, opened 2016-11-09
soft delete video for a course

muhammad-ammar:
### Description

Add support to soft delete a video for a particular course.

### Testing
- [x] Unit

### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [ ] Code review: @muzaffaryousaf 
- [x] Code review: @mushtaqak 

### Post-review
- [ ] Rebase and squash commits

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -147,7 +147,7 @@ def update_video(video_data):
     try:
         video = _get_video(video_data.get("edx_video_id"))
     except Video.DoesNotExist:
-        error_message = u"Video not found when trying to update video with edx_video_id: {0}".format(video_data.get("edx_video_id")) 
+        error_message = u"Video not found when trying to update video with edx_video_id: {0}".format(video_data.get("edx_video_id"))
         raise ValVideoNotFoundError(error_message)
 
     serializer = VideoSerializer(video, data=video_data)
@@ -328,10 +328,23 @@ def get_videos_for_course(
         total order.
     """
     return _get_videos_for_filter(
-        {"courses__course_id":unicode(course_id)},
+        {"courses__course_id": unicode(course_id), "courses__is_hidden": False},
         sort_field,
         sort_dir,
     )
+
+
+def remove_video_for_course(course_id, edx_video_id):
+    """
+    Soft deletes video for particular course.
+
+    Arguments:
+        course_id (str): id of the course
+        edx_video_id (str): id of the video to be hidden
+    """
+    course_video = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id)
+    course_video.is_hidden = True
+    course_video.save()
 
 
 def get_videos_for_ids(
--- a/edxval/models.py
+++ b/edxval/models.py
@@ -125,6 +125,7 @@ class CourseVideo(models.Model, ModelFactoryWithValidation):
     """
     course_id = models.CharField(max_length=255)
     video = models.ForeignKey(Video, related_name='courses')
+    is_hidden = models.BooleanField(default=False, help_text='Hide video for course.')
 
     class Meta:  # pylint: disable=C1001
         """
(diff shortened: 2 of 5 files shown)


edxval/tests/test_api.py
```diff
@@ -1055,3 +1055,47 @@ def test_invalid_encoded_video_attr(self):
     def test_invalid_course_id(self):
         xml = self.make_import_xml(video_dict=constants.VIDEO_DICT_FISH)
         self.assert_invalid_import(xml, "x" * 300)
+
+
+class GetCourseVideoRemoveTest(TestCase):
+    """
+    Tests for our remove_video_for_course function in api.py
```
mushtaqak: I would have written `Tests to check 'remove_video_for_course' method works correctly.` or something like that.

edxval/tests/test_api.py
```diff
@@ -1055,3 +1055,47 @@ def test_invalid_encoded_video_attr(self):
...
+    """
+
+    def setUp(self):
+        """
+        Creates Video objects in database
```
mushtaqak: I would have written it as `Creates video objects for courses` or something like that.

muhammad-ammar:
@mushtaqak Added your suggestions.

edxval/api.py
```diff
@@ -328,12 +328,25 @@ def get_videos_for_course(
...
-        {"courses__course_id":unicode(course_id)},
+        {"courses__course_id": unicode(course_id), "courses__is_hidden": False},
         sort_field,
         sort_dir,
     )
 
 
+def remove_video_for_course(course_id, edx_video_id):
+    """
+    Remove video for a particular course.
```
muzaffaryousaf: `Soft deletes video for particular course.`

edxval/models.py
```diff
@@ -125,6 +125,7 @@ class CourseVideo(models.Model, ModelFactoryWithValidation):
     """
     course_id = models.CharField(max_length=255)
     video = models.ForeignKey(Video, related_name='courses')
+    is_hidden = models.BooleanField(default=False, help_text='Hide this video for course.')
```
muzaffaryousaf: `Hide video for course.`

edxval/tests/test_api.py
```diff
@@ -1055,3 +1055,47 @@ def test_invalid_encoded_video_attr(self):
...
+        """
+        Creates video objects for courses
+        """
+        # create video in the test course
+        self.course_id = 'test-course'
+        video = Video.objects.create(**constants.VIDEO_DICT_FISH)
+        CourseVideo.objects.create(video=video, course_id=self.course_id)
+        self.edx_video_id = video.edx_video_id
+
+        # add the video in another course (to make sure that video is remove for correct course )
```
muzaffaryousaf: `remove` --> `removed`

muzaffaryousaf:
@muhammad-ammar This LGTM 👍 except couple of suggestions.

muzaffaryousaf:
Squash & merge.

muhammad-ammar:
We are not handling the the exception in case video doesn't exist. So i am
not sure if we really need to write a test for this case.

BenjiLee:
In that case, should we be handling that exception?

What happens when you try to delete a video that doesn't exist? 404? 500? 200?

muhammad-ammar:
@BenjiLee We get `DoesNotExist` exception when video doesn't exist. In
edxval at other places we catch this exception and then raised another
exception which is `ValVideoNotFoundError`. But in video removal case i am
not raising `ValVideoNotFoundError`. We are using edxval as an internal api
from within edx-platform without any HTTP REST interface so there is no
HTTP status code and we will get an unhandled exception whether we raise
`DoesNotExist` or `ValVideoNotFoundError`.

We are deleting videos from studio `Video Uploads` page so this case(video
doesn't exist) will not happen. Even if for some reason this happens,
studio will show a default error popup.

Do you see any issue with current approach?

BenjiLee:
Ah, didn't realize right away it was an internal call. Good to go.

PR #65, opened 2016-11-09
Increment version to 0.0.10

BenjiLee:
FYI @jagonzalr


BenjiLee:
@jagonzalr Version bumped in https://github.com/edx/edx-val/pull/64

PR #66, opened 2016-12-06
Fix course video migration

mushtaqak:
# Description

This PR fixes the migration that was missed in [PR#64](https://github.com/edx/edx-val/pull/64)

### [TNL-6094](https://openedx.atlassian.net/browse/TNL-6094)

@muzaffaryousaf quick review ?


muzaffaryousaf:
Let's merge this 👍 and make a PR in platform too.

PR #67, opened 2016-12-22
update video status

muhammad-ammar:
[TNL-4777](https://openedx.atlassian.net/browse/TNL-4777)

@mushtaqak This is ready for review.

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -158,6 +158,30 @@ def update_video(video_data):
         raise ValCannotUpdateError(serializer.errors)
 
 
+def update_video_status(edx_video_id, status):
+    """
+    Update status for an existing video.
+
+    Args:
+        edx_video_id: ID of the video
+        status: video status
+
+    Raises:
+        Raises ValVideoNotFoundError if the video cannot be retrieved.
+    """
+
+    try:
+        video = _get_video(edx_video_id)
+    except Video.DoesNotExist:
+        error_message = u"Video not found when trying to update video status with edx_video_id: {0}".format(
+            edx_video_id
+        )
+        raise ValVideoNotFoundError(error_message)
+
+    video.status = status
+    video.save()
+
+
 def create_profile(profile_name):
     """
     Used to create Profile objects in the database
--- a/edxval/models.py
+++ b/edxval/models.py
@@ -11,10 +11,14 @@ themselves. After these are resolved, errors such as a negative file_size or
 invalid profile_name will be returned.
 """
 
+import logging
+
 from django.db import models
+from django.dispatch import receiver
 from django.core.validators import MinValueValidator, RegexValidator
 from django.core.urlresolvers import reverse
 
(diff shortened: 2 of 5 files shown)


mushtaqak:
LGTM 👍

muhammad-ammar:
@cahrens Do you have some time to review this PR?

cahrens:
@muhammad-ammar I'm out sick today, and then on vacation until after the New Year (and I don't actually have any specific knowledge of video state). I'd suggest that you post in the TNL hipchat room for a reviewer.

muhammad-ammar:
@cahrens oh..get well soon

edxval/api.py
```diff
@@ -158,6 +158,33 @@ def update_video(video_data):
         raise ValCannotUpdateError(serializer.errors)
 
 
+def update_video_status(video_data):
```
nasthagiri: Nit: Per clean-code best practices, it's better to send only the data that this function needs as separate parameters.  So `video_id` and `status` should be separate parameters to this function.  And then, the implementation of this method need not be as defensive - that is, no longer a need to call `get` on the given `dict`.

edxval/models.py
```diff
@@ -190,3 +194,16 @@ def content_type(self):
...
+
+@receiver(models.signals.post_save, sender=Video)
+def video_status_update_callback(sender, **kwargs):  # pylint: disable=unused-argument
+    """
+    Log video status for an existing video instance
+    """
+    video = kwargs['instance']
+
+    if kwargs['created']:
+        logger.info('Video created with id [%s] and status [%s]', video.edx_video_id, video.status)
```
nasthagiri: Nit: I find it very useful to have a common prefix for all Splunk logs within a module.  It then makes it easier to query in Splunk.  So, for example, for VAL, you may consider prefixing all log messages with "VAL: "

edxval/tests/test_api.py
```diff
@@ -1099,3 +1100,22 @@ def test_remove_video_for_course(self):
...
+    """
+    @patch('edxval.models.logger')
+    def test_update_video_status_logging(self, mock_logger):
+        """
+        Tests video status update logging
+        """
+        video = Video.objects.create(**constants.VIDEO_DICT_FISH)
+        mock_logger.info.assert_called_with(
+            'Video created with id [%s] and status [%s]', video.edx_video_id, constants.VIDEO_DICT_FISH.get('status')
+        )
```
nasthagiri: You may want to also have a negative test here - to make sure the "Status changed to" logging does *not* happen at the time the `Video` object is created.

muhammad-ammar: @nasthagiri But we are already verifying that correct message is logged upon `Video` object creation. If for some reason incorrect message is logged when `Video` object is created then the above assert will surely fail. Am i missing something?

muhammad-ammar:
@nasthagiri Addressed your feedback.

PR #71, opened 2017-04-06
Add course video image upload support

mushtaqak:


edxval/api.py
```diff
@@ -182,6 +183,35 @@ def update_video_status(edx_video_id, status):
     video.save()
 
 
+def update_video_image(edx_video_id, course_id, serialized_data):
+    """
+    Update video image for an existing video.
+
+    Args:
+        edx_video_id: ID of the video
+        serialized_data: Serialized data of image for video.
```
mushtaqak: course_id is missed, i will add it in next code push

edxval/views.py
```diff
@@ -56,8 +56,8 @@ class VideoList(generics.ListCreateAPIView):
     """
     GETs or POST video objects
     """
-    authentication_classes = (OAuth2Authentication, SessionAuthentication)
-    permission_classes = (ReadRestrictedDjangoModelPermissions,)
+    # authentication_classes = (OAuth2Authentication, SessionAuthentication)
```
mushtaqak: I will remove these in next code push

edxval/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+from django.conf import settings
+from django.core.files.storage import get_storage_class
+
+
+def video_image_path(video_image_instance, filename):  # pylint:disable=unused-argument
+    """
+    Returns video image path.
+    """
+    return '{}{}'.format(settings.VIDEO_IMAGE_SETTINGS.get('DIRECTORY_PREFIX', ''), filename)
```
muzaffaryousaf: Why `video_image_instance` ?

edxval/serializers.py
```diff
@@ -122,6 +124,21 @@ class Meta:  # pylint: disable=C1001, C0111
         lookup_field = "edx_video_id"
         exclude = ('id',)
 
+    def get_course_video_image_url(self, video):
```
Qubad786: Rather than doing this in `VideoSerializer`, we should create `CourseVideoSerializer` (and use it in api), since we intend to use course-specific video thumbnail url as `VideoImage` is having in one to one relation with `CourseVideo`.

edxval/api.py
```diff
@@ -355,6 +356,7 @@ def get_videos_for_course(
         {"courses__course_id": unicode(course_id), "courses__is_hidden": False},
         sort_field,
         sort_dir,
+        context={'course_id': course_id}
```
Qubad786: We should show consistency i.e. use either single/double quote strings throughout.

edxval/models.py
```diff
@@ -11,13 +11,19 @@
 invalid profile_name will be returned.
 """
 
+from contextlib import closing
 import logging
+import os
+
+from model_utils.models import TimeStampedModel
```
Qubad786: Import order: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#imports-order

edxval/settings.py
```diff
@@ -178,3 +179,12 @@
...
 COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
 COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
+
+VIDEO_IMAGE_SETTINGS = dict(
+    # Backend storage
+    # STORAGE_CLASS='storages.backends.s3boto.S3BotoStorage',
+    # STORAGE_KWARGS=dict(bucket='video-image-test-bucket'),
+    # If you are changing prefix value then update the .gitignore accordingly
+    # so that images created during tests due to upload should be ignored
+    DIRECTORY_PREFIX='videoimage/',
```
Qubad786: missing a preceding `/` i.e. `'/video-image/'`

edxval/api.py
```diff
@@ -1,69 +1,31 @@
 # pylint: disable=E1101
 # -*- coding: utf-8 -*-
 """
-The internal API for VAL. This is not yet stable
```
Qubad786: is there any context behind `This is not yet stable` which has been removed from L4

edxval/utils.py
muhammad-ammar: `CustomizableImageField` passes the `VideoImage` model instance as first param.

edxval/api.py
```diff
@@ -182,6 +144,53 @@ def update_video_status(edx_video_id, status):
...
+
+    Returns:
+        course video image url
+    """
+    storage = get_video_image_storage()
+
+    try:
+        if video_image is None:
+            video_image = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id).video_image
+    except ObjectDoesNotExist:
```
mushtaqak: I think we should avoid adding extra import `ObjectDoesNotExist` and use `CourseVideo.DoesNotExist`.

edxval/api.py
mushtaqak: using single quotes is better, we should always use them whenever we can.  May be update above quotes.

edxval/exceptions.py
```diff
@@ -0,0 +1,57 @@
...
+class ValCannotUpdateError(ValError):
+    """
+    This error is raised when an object cannot be updated
+    """
+    pass
+
+
+class ValVideoImageNotFoundError(ValError):
+    """
+    This error is raised when a video image is not found
```
mushtaqak: a period (dot) at the end of the docstring ?

edxval/api.py
muhammad-ammar: I don't think this line makes sense and API is stable enough.

edxval/serializers.py
muhammad-ammar: As per discussion with @muzaffaryousaf we will keep the implementation as it is. We will later update it once the design document is completed.

edxval/models.py
```diff
@@ -72,6 +78,32 @@ def __unicode__(self):
...
+    be serialized (hard-coded) in migrations. Otherwise,
+    migrations include optional settings for storage (such as
+    the storage class and bucket name); we don't want to
+    create new migration files for each configuration change.
+    """
+    def __init__(self, *args, **kwargs):
+        kwargs.update(dict(
+            upload_to=video_image_path,
+            storage=get_video_image_storage(),
+            max_length=500,  # allocate enough for filepath
```
muhammad-ammar: I think this is large enough.

muhammad-ammar:
@muzaffaryousaf @Qubad786 @mushtaqak Feedback addressed.

muhammad-ammar:
@muzaffaryousaf @Qubad786 @mushtaqak I have updated the way course video image is serialised. Please review.

.gitignore
```diff
@@ -66,3 +66,5 @@ logs/*/*.log*
 .vagrant
 
 venv/
+
+video-image/
```
mushtaqak: This needs to be `video-images/` now

edxval/api.py
```diff
@@ -182,6 +144,53 @@ def update_video_status(edx_video_id, status):
...
+        file_name: File name of the image file.
+
+    Raises:
+        Raises ValVideoNotFoundError if the CourseVideo cannot be retrieved.
+        Raises ValVideoImageNotFoundError if the VideoImage cannot be retrieved.
+    """
+    try:
+        course_video = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id)
+    except Video.DoesNotExist:
+        error_message = u"CourseVideo not found for edx_video_id: {0}".format(edx_video_id)
```
mushtaqak: I would use single quotes here?

edxval/settings.py
```diff
@@ -178,3 +179,12 @@
 # copied from edx-platform
 COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/]+)'
 COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
+
+VIDEO_IMAGE_SETTINGS = dict(
+    # Backend storage
```
mushtaqak: I think we need to add max and min size, as used in config PR.
```
VIDEO_IMAGE_MAX_BYTES : 2097152
VIDEO_IMAGE_MIN_BYTES : 100 
```

edxval/settings.py
```diff
@@ -178,3 +179,12 @@
...
+    # STORAGE_CLASS='storages.backends.s3boto.S3BotoStorage',
+    # STORAGE_KWARGS=dict(bucket='video-image-bucket'),
+    # If you are changing prefix value then update the .gitignore accordingly
+    # so that images created during tests due to upload should be ignored
+    DIRECTORY_PREFIX='video-image/',
```
mushtaqak: This would change to `video-images/`

requirements.txt
```diff
@@ -4,3 +4,7 @@ enum34==1.0.4
 lxml==3.3.6
 -e git+https://github.com/edx/django-oauth2-provider.git@0.2.7-fork-edx-6a#egg=django-oauth2-provider==0.2.7-fork-edx-6
 -e git+https://github.com/edx/django-rest-framework-oauth.git@f0b503fda8c254a38f97fef802ded4f5fe367f7a#egg=djangorestframework-oauth
+django-storages==1.5.2
+boto==2.46.1
+Pillow==4.1.0
```
mushtaqak: Do we need Pillow now ?

edxval/utils.py
```diff
@@ -0,0 +1,25 @@
...
+def get_video_image_storage():
+    """
+    Return the configured django storage backend.
+    """
+    if hasattr(settings, 'VIDEO_IMAGE_SETTINGS'):
+        return get_storage_class(
+            settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS'),
+        )(**settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_KWARGS', {}))
+    else:
+        return get_storage_class()
```
mushtaqak: This needs a comment describing why and what this does.

edxval/admin.py
```diff
@@ -30,6 +33,13 @@ class VideoAdmin(admin.ModelAdmin):  # pylint: disable=C0111
...
+
+class VideoImageAdmin(admin.ModelAdmin):
+    model = VideoImage
+    verbose_name = "Video Image"
+    verbose_name_plural = "Video Images"
+
 admin.site.register(Profile, ProfileAdmin)
 admin.site.register(Video, VideoAdmin)
 admin.site.register(Subtitle)
+admin.site.register(VideoImage, VideoImageAdmin)
```
muzaffaryousaf: can we also add CourseVideo to the admin in this PR or any other/subsequent PR ?

edxval/admin.py
```diff
@@ -30,6 +33,13 @@ class VideoAdmin(admin.ModelAdmin):  # pylint: disable=C0111
     admin_order_field = 'edx_video_id'
     inlines = [CourseVideoInline, EncodedVideoInline]
 
+
+class VideoImageAdmin(admin.ModelAdmin):
+    model = VideoImage
+    verbose_name = "Video Image"
+    verbose_name_plural = "Video Images"
```
muzaffaryousaf: Let's use single quotes ?

edxval/utils.py
```diff
@@ -0,0 +1,27 @@
...
+Util methods to be used in api and models.
+"""
+
+from django.conf import settings
+from django.core.files.storage import get_storage_class
+
+
+def video_image_path(video_image_instance, filename):  # pylint:disable=unused-argument
+    """
+    Returns video image path.
```
muzaffaryousaf: Let's add Arguments in this doc string and expectation about video_image_instance since it's not being used in this method which can confuse readers.

edxval/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    return '{}{}'.format(settings.VIDEO_IMAGE_SETTINGS.get('DIRECTORY_PREFIX', ''), filename)
```
muzaffaryousaf: I would use `u` along with format. This has beaten us many time at different places.

edxval/api.py
muzaffaryousaf: Infact we need to catch two exceptions: 1 for CourseVideo.DoesNotExist and 1 for course_video.video_image.

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
+    Returns:
+        course video image url or None if no image found
+    """
+    storage = get_video_image_storage()
+
+    try:
+        if video_image is None:
+            video_image = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id).video_image
+    except ObjectDoesNotExist:
+        return None
```
muzaffaryousaf: Do we need to log the error here ?

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
+
+    return storage.url(video_image.image.name)
```
muzaffaryousaf: what if name is empty, is it okay to call storage.url ?

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
+
+def update_video_image(edx_video_id, course_id, image_data, file_name):
+    """
+    Update video image for an existing video.
+
+    Arguments:
+        edx_video_id: ID of the video.
+        course_id: ID of course.
+        image_data: Image data for video.
+        file_name: File name of the image file.
```
muzaffaryousaf: I think the argument description is too obvious and not providing any extra context here. Better to remove them ? We can leave image_data as it by adding the type (InMemoryUploadedFile).

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
+
+    Raises:
+        Raises ValVideoNotFoundError if the CourseVideo cannot be retrieved.
+    """
+    try:
+        course_video = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id)
+    except Video.DoesNotExist:
```
muzaffaryousaf: Should't  this catch DoesNotExist for both tables i.e. Video & CourseVideo ?

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
 
 
+def get_course_video_image_url(course_id=None, edx_video_id=None, video_image=None):
+    """
+    Returns course video image url.
+
+    Arguments:
+        course_id: ID of course
+        edx_video_id: ID of the video
+        video_image: VideoImage instance
```
muzaffaryousaf: Too obvious description for argument wither we need update or simply remove it.

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
     video.save()
 
 
+def get_course_video_image_url(course_id=None, edx_video_id=None, video_image=None):
```
muzaffaryousaf: Why is everything optional ?

muzaffaryousaf: We're calling `video_image` after `VideoImage.create_or_update` so why there would be None in video_image ?

edxval/models.py
```diff
@@ -155,6 +163,76 @@ class EncodedVideo(models.Model):
...
+        """
+        video_image, created = cls.objects.get_or_create(course_video=course_video)
+        if image_data:
+            with closing(image_data) as image_file:
+                file_name = '{uuid}{ext}'.format(uuid=uuid4().hex, ext=os.path.splitext(file_name)[1])
+                video_image.image.save(file_name, image_file)
+        else:
+            video_image.image.name = file_name
+
+        video_image.save()
```
muzaffaryousaf: I think we should enclose this in exception and log it for debug purposes.

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
+        image_data: Image data for video.
+        file_name: File name of the image file.
+
+    Raises:
+        Raises ValVideoNotFoundError if the CourseVideo cannot be retrieved.
+    """
+    try:
+        course_video = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id)
+    except Video.DoesNotExist:
+        error_message = u'CourseVideo not found for edx_video_id: {0}'.format(edx_video_id)
```
muzaffaryousaf: The exception is capturing Video but in error message we're writing about CourseVideo that does't sound okay to me.

edxval/api.py
```diff
@@ -182,6 +143,52 @@ def update_video_status(edx_video_id, status):
...
+    Update video image for an existing video.
+
+    Arguments:
+        edx_video_id: ID of the video.
+        course_id: ID of course.
+        image_data: Image data for video.
+        file_name: File name of the image file.
+
+    Raises:
+        Raises ValVideoNotFoundError if the CourseVideo cannot be retrieved.
```
muzaffaryousaf: Returns:
        course video image url in this <format>

edxval/models.py
```diff
@@ -155,6 +163,76 @@ class EncodedVideo(models.Model):
...
+            image_data (InMemoryUploadedFile): Image data to be saved.
+
+        Returns:
+            Returns a tuple of (video_image, created).
+        """
+        video_image, created = cls.objects.get_or_create(course_video=course_video)
+        if image_data:
+            with closing(image_data) as image_file:
+                file_name = '{uuid}{ext}'.format(uuid=uuid4().hex, ext=os.path.splitext(file_name)[1])
+                video_image.image.save(file_name, image_file)
```
muzaffaryousaf: We need to enclose such things in exception and logg those exception in case of errors.

edxval/api.py
muzaffaryousaf: Can't we use `image_url` defined in models.py instead of this

edxval/api.py
muhammad-ammar: I don't think we need logging here.

muhammad-ammar:
@muzaffaryousaf Addressed your feedback.

edxval/api.py
```diff
@@ -182,6 +142,42 @@ def update_video_status(edx_video_id, status):
     video.save()
 
 
+def get_course_video_image_url(course_id=None, edx_video_id=None, video_image=None):
```
Qubad786: what's the purpose of calling it without the arguments if all are optional?

edxval/api.py
```diff
@@ -182,6 +142,42 @@ def update_video_status(edx_video_id, status):
...
+
+    return video_image.image_url()
+
+
+def update_video_image(edx_video_id, course_id, image_data, file_name):
+    """
+    Update video image for an existing video.
+
+    Arguments:
+        image_data (InMemoryUploadedFile): Image data to be saved for a course video.
```
Qubad786: Please consider adding all the arguments into the doc string.

muzaffaryousaf: As discussed, we don't need to add rest of the args.

edxval/admin.py
muzaffaryousaf: @muhammad-ammar left by mistake ?

edxval/models.py
```diff
@@ -137,6 +145,15 @@ class Meta:  # pylint: disable=C1001
         """
         unique_together = ("course_id", "video")
 
+    def image_url(self):
+        """
+        Return image url for a course video image.
+        """
+        try:
+            return self.video_image.image_url()
```
muzaffaryousaf: can we use hasattr here to not use try except, something like
```
if hassattr(self, 'video_image'):
    self.video_image.image_url()
```

edxval/admin.py
muhammad-ammar: I will do this in a separate PR.

edxval/models.py
```diff
@@ -147,12 +147,10 @@ class Meta:  # pylint: disable=C1001
 
     def image_url(self):
         """
-        Return image url for a course video image.
+        Return image url for a course video image or None if no image.
```
Qubad786: > Return image url for a course video image or None if no image.

Its not returning None if `hasattr(self, 'video_image')` is `False`

muhammad-ammar: By default a method returns `None` if nothing is returned.

edxval/api.py
```diff
@@ -182,6 +143,47 @@ def update_video_status(edx_video_id, status):
...
+    try:
+        video_image = CourseVideo.objects.select_related('video_image').get(
+            course_id=course_id, video__edx_video_id=edx_video_id
+        ).video_image
+        return video_image.image_url()
+    except ObjectDoesNotExist:
+        return None
+
+
+def update_video_image(edx_video_id, course_id, image_data, file_name):
```
Qubad786: We need to specify in the doc string about why do we need filename because an obvious thought that comes into mind is; this filename would be used as it is for the image but it is only needed to decide the image's extension.

edxval/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    Return the configured django storage backend.
+    """
+    if hasattr(settings, 'VIDEO_IMAGE_SETTINGS'):
+        return get_storage_class(
+            settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS'),
+        )(**settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_KWARGS', {}))
+    else:
+        # during edx-platform loading this method gets called but settings are not ready yet
+        # so in that case we will return default(FileSystemStorage) storage class instance
+        return get_storage_class()()
```
Qubad786: We should double check that we do not end up storing the images in `FileSystemStorage` due to this default  storage set on platform load.

edxval/models.py
```diff
@@ -155,6 +170,84 @@ class EncodedVideo(models.Model):
...
+        Returns:
+            Returns a tuple of (video_image, created).
+        """
+        video_image, created = cls.objects.get_or_create(course_video=course_video)
+        if image_data:
+            with closing(image_data) as image_file:
+                file_name = '{uuid}{ext}'.format(uuid=uuid4().hex, ext=os.path.splitext(file_name)[1])
+                try:
+                    course_id = course_video.course_id
+                    edx_video_id = course_video.video.edx_video_id
```
muzaffaryousaf: If we need these in exception then we should move these under exception.

edxval/models.py
```diff
@@ -155,6 +170,84 @@ class EncodedVideo(models.Model):
...
+                    video_image.image.save(file_name, image_file)
+                except Exception:  # pylint: disable=broad-except
+                    logger.exception(
+                        'VAL: Video Image save failed to storage for course_id [%s] and video_id [%s]',
+                        course_id,
+                        edx_video_id
+                    )
+                    raise
+        else:
+            video_image.image.name = file_name
```
muzaffaryousaf: Are we using user specified `file_name`  here ? I thought we'll store uuid always as image name.

edxval/models.py
```diff
@@ -155,6 +170,84 @@ class EncodedVideo(models.Model):
...
+    image = CustomizableImageField()
+
+    @classmethod
+    def create_or_update(cls, course_video, file_name, image_data=None):
+        """
+        Create a VideoImage object for a CourseVideo.
+
+        Arguments:
+            course_video (CourseVideo): CourseVideo instance
+            file_name (str): File name of the image
```
muzaffaryousaf: I think in few cases this will be user-specified name and in some cases this will be existing name, can we make it clear in doc strings ?

edxval/serializers.py
```diff
@@ -168,9 +181,12 @@ def create(self, validated_data):
 
         # The CourseSerializer will already have converted the course data
         # to CourseVideo models, so we can just set the video and save.
-        for course_video in courses:
+        # Also create VideoImage objects if an image filename is present
+        for course_video, image in courses:
```
muzaffaryousaf: image is not really image here. can we rename it to something more readable ?

edxval/models.py
muzaffaryousaf: it's clear to me now.

muhammad-ammar:
@muzaffaryousaf @Qubad786 Feedback addressed.

PR #72, opened 2017-05-08
support video images for course re-run and import/export

muhammad-ammar:


muhammad-ammar:
@muzaffaryousaf @mushtaqak @Qubad786 This is ready for review.

muzaffaryousaf:
@mushtaqak  @Qubad786 can you guys take a look first, i'll probably do a final quick check.

.travis.yml
```diff
@@ -10,5 +10,6 @@ script:
 branches:
     only:
       - master
+      - ammar/course-rerun-import-export
```
mushtaqak: Why do we need this ?

edxval/api.py
```diff
@@ -491,34 +492,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+            video=course_video.video,
             course_id=destination_course_id
         )
+        try:
+            VideoImage.create_or_update(
+                course_video=dest_course_video,
+                file_name=course_video.video_image.image.name
+            )
+        except VideoImage.DoesNotExist:
+            pass
```
mushtaqak: Would we ever get the exception here? I mean, we are using `create_or_update` so if not exist it will create, otherwise update. What is the best approach here?

edxval/api.py
```diff
@@ -491,34 +492,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
         ValVideoNotFoundError: if the video does not exist
     """
+    image = ''
     video = _get_video(edx_video_id)
+
+    try:
+        course_video = CourseVideo.objects.select_related('video_image').get(course_id=course_id, video=video)
+        image = course_video.video_image.image.name
+    except ObjectDoesNotExist:
+        pass
```
mushtaqak: Do we need to add some logging here?

edxval/tests/test_api.py
```diff
@@ -807,16 +809,21 @@ def setUp(self):
 
     def test_successful_copy(self):
         """Tests a successful copy course"""
-        api.copy_course_videos('test-course', 'course-copy1')
+        destination_course_id = 'course-copy1'
+        api.copy_course_videos('test-course', destination_course_id)
```
mushtaqak: I think we can use `self.course_id` for `test-course`

edxval/tests/test_api.py
```diff
@@ -899,23 +910,29 @@ def parse_xml(self, xml_str):
...
                 <encoded_video url="http://www.meowmix.com" file_size="11" bitrate="22" profile="mobile"/>
                 <encoded_video url="http://www.meowmagic.com" file_size="33" bitrate="44" profile="desktop"/>
                 <encoded_video url="https://www.tmnt.com/tmnt101.m3u8" file_size="100" bitrate="0" profile="hls"/>
             </video_asset>
-        """)
+        """.format(image=image))
+
         self.assert_xml_equal(
-            api.export_to_xml(constants.VIDEO_DICT_FISH["edx_video_id"]),
+            api.export_to_xml(constants.VIDEO_DICT_FISH["edx_video_id"], course_id),
```
mushtaqak: I would use single quotes.

.travis.yml
muhammad-ammar: so that tests in travis can be run on this branch. otherwise tests will run only on master branch.

edxval/api.py
muhammad-ammar: The exception will be raised on `course_video.video_image.image.name` if source course video does not have a related video image.

edxval/api.py
muhammad-ammar: I am not sure if we really need logging here.

edxval/api.py
```diff
@@ -489,34 +490,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
-    for video in videos:
-        CourseVideo.objects.get_or_create(
-            video=video,
+    for course_video in course_videos:
+        dest_course_video, __ = CourseVideo.objects.get_or_create(
+            video=course_video.video,
             course_id=destination_course_id
         )
+        try:
+            VideoImage.create_or_update(
```
Qubad786: Do we need to do this in an atomic transaction covering creation if `dest_course_video` too? what are our directions if `VideoImage.create_or_update` for some reason by that time `dest_course_video` would've already been created. Do we need rollback in that case?

edxval/api.py
```diff
@@ -561,7 +581,12 @@ def import_from_xml(xml, edx_video_id, course_id=None):
             course_id,
         )
         if course_id:
-            CourseVideo.get_or_create_with_validation(video=video, course_id=course_id)
+            course_video, __ = CourseVideo.get_or_create_with_validation(video=video, course_id=course_id)
+
+            image_file_name = xml.get('image', '').strip()
+            if image_file_name:
+                VideoImage.create_or_update(course_video, image_file_name)
```
Qubad786: Same as above, Do we need atomic transaction covering creation of `CourseVideo` and `VideoImage` here?

edxval/api.py
```diff
@@ -489,34 +490,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
-    videos = Video.objects.filter(courses__course_id=unicode(source_course_id))
+    course_videos = CourseVideo.objects.select_related('video', 'video_image').filter(
+        course_id=unicode(source_course_id)
+    )
 
-    for video in videos:
-        CourseVideo.objects.get_or_create(
-            video=video,
+    for course_video in course_videos:
+        dest_course_video, __ = CourseVideo.objects.get_or_create(
```
muzaffaryousaf: `destination_course_video`

edxval/api.py
```diff
@@ -489,34 +490,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
         edx_video_id (str): The ID of the video to export
+        course_id (str): The ID of the course with which this video is associated
 
     Returns:
         An lxml video_asset element containing export data
 
     Raises:
         ValVideoNotFoundError: if the video does not exist
     """
+    image = ''
```
muzaffaryousaf: is't it `video_image_name` ?

edxval/api.py
```diff
@@ -576,7 +601,7 @@ def import_from_xml(xml, edx_video_id, course_id=None):
         'duration': xml.get('duration'),
         'status': 'imported',
         'encoded_videos': [],
-        'courses': [course_id] if course_id else [],
+        'courses': [{course_id: xml.get('image')}] if course_id else [],
```
muzaffaryousaf: Rough guess --> i believe we're not adding any backward compatibility here. Let's say someone tries to import already exported course which don't have image property, will that work ?

edxval/tests/test_api.py
```diff
@@ -807,16 +809,21 @@ def setUp(self):
 
     def test_successful_copy(self):
         """Tests a successful copy course"""
```
muzaffaryousaf: Can we change this too ?

edxval/tests/test_api.py
```diff
@@ -938,14 +956,18 @@ def setUp(self):
...
         ret = etree.Element(
             "video_asset",
             attrib={
                 key: unicode(video_dict[key])
                 for key in ["client_video_id", "duration"]
             }
         )
+
+        if image:
+            ret.attrib['image'] = image
```
muzaffaryousaf: I would rename `ret` to something more useful for better readability.

edxval/api.py
```diff
@@ -489,34 +490,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
+    for course_video in course_videos:
+        dest_course_video, __ = CourseVideo.objects.get_or_create(
+            video=course_video.video,
             course_id=destination_course_id
         )
+        try:
+            VideoImage.create_or_update(
+                course_video=dest_course_video,
+                file_name=course_video.video_image.image.name
+            )
```
muzaffaryousaf: Is it possible to write few tests around number of queries that will executed ?

edxval/api.py
muhammad-ammar: This will work. `xml.get('image')` will return `None` if `image` attribute is not present in `xml`.

edxval/api.py
muhammad-ammar: I don't think we need atomic transactions in the above two cases.

muhammad-ammar:
@Qubad786 @muzaffaryousaf Feedback addressed.

edxval/api.py
```diff
@@ -148,7 +149,9 @@ def get_course_video_image_url(course_id=None, edx_video_id=None, video_image=No
     """
     try:
         if video_image is None:
-            video_image = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id).video_image
+            video_image = CourseVideo.objects.select_related('video_image').get(
+                course_id=course_id, video__edx_video_id=edx_video_id
+            ).video_image
```
muzaffaryousaf: Here we're doing 2 queries to the database. 1 to get course video object via `.get()` and second on video image via `.video_image`. My question is that ObjectDoesNotExist will capture both of them and do we have test coverage for that ?

muzaffaryousaf: In either case i think we should be specific about DoesNotExist based on it's model.

edxval/api.py
```diff
@@ -169,7 +172,9 @@ def update_video_image(edx_video_id, course_id, image_data, file_name):
         Raises ValVideoNotFoundError if the CourseVideo cannot be retrieved.
     """
     try:
-        course_video = CourseVideo.objects.get(course_id=course_id, video__edx_video_id=edx_video_id)
+        course_video = CourseVideo.objects.select_related('video').get(
+            course_id=course_id, video__edx_video_id=edx_video_id
+        )
     except ObjectDoesNotExist:
         error_message = u'CourseVideo not found for edx_video_id: {0}'.format(edx_video_id)
```
muzaffaryousaf: do we also need to log course id ?

edxval/api.py
```diff
@@ -478,34 +483,53 @@ def copy_course_videos(source_course_id, destination_course_id):
...
-            video=video,
+    for course_video in course_videos:
+        destination_course_video, __ = CourseVideo.objects.get_or_create(
+            video=course_video.video,
             course_id=destination_course_id
         )
+        try:
+            VideoImage.create_or_update(
+                course_video=destination_course_video,
+                file_name=course_video.video_image.image.name
```
Qubad786: more readable if we check hasattr(course_video, 'video_image') rather than catching does not exist.

muhammad-ammar:
@Qubad786 @muzaffaryousaf Feedback addressed. I am merging this.

PR #73, opened 2017-05-18
add support in model for multiple video image urls

muhammad-ammar:
@muzaffaryousaf @mushtaqak @Qubad786 This is ready for review.

### DIFF
--- a/edxval/models.py
+++ b/edxval/models.py
@@ -12,12 +12,14 @@ invalid profile_name will be returned.
 """
 
 from contextlib import closing
+import json
 import logging
 import os
 from uuid import uuid4
 
 from django.db import models
 from django.dispatch import receiver
+from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator, RegexValidator
 from django.core.urlresolvers import reverse
 
@@ -28,6 +30,7 @@ from edxval.utils import video_image_path, get_video_image_storage
 logger = logging.getLogger(__name__)  # pylint: disable=C0103
 
 URL_REGEX = r'^[a-zA-Z0-9\-_]*$'
+LIST_MAX_ITEMS = 3
 
 
 class ModelFactoryWithValidation(object):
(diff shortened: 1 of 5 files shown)


edxval/tests/test_api.py
```diff
@@ -1335,3 +1335,30 @@ def test_update_video_image_exception(self):
...
+        course_video = CourseVideo.objects.create(video=self.video, course_id='course101')
+        video_image = VideoImage.objects.create(course_video=course_video)
+        video_image.urls = image_urls
+        video_image.save()
+        self.assertEqual(video_image.urls, image_urls)
+
+        # expect a validation error if we try to set non list data
+        exception_messages = set()
+        for item in ('a string', 555, {'a': 1}, (1,), video_image):
+            with self.assertRaises(Exception) as set_exception:
```
Qubad786: We should be asserting `ValidationError` specifically.

edxval/tests/test_api.py
```diff
@@ -1335,3 +1335,30 @@ def test_update_video_image_exception(self):
...
+        """
+        Test `VideoImage.urls` field works as expected.
+        """
+        image_urls = ['video-images/a.png', 'video-images/b.png']
+
+        # an empty list should be returned when there is no value for urls
+        self.assertEqual(self.course_video.video_image.urls, [])
+
+        # set a list with data and expect the same list to be returned
+        course_video = CourseVideo.objects.create(video=self.video, course_id='course101')
```
Qubad786: Can we also assert that we get same result when access the `urls` via reverse relation.

edxval/tests/test_api.py
```diff
@@ -1335,3 +1335,30 @@ def test_update_video_image_exception(self):
                 does_not_course_id
             )
         )
+
+    def test_video_image_urls_field(self):
```
Qubad786: This should be moved to `test_models.py`, since its testing the model's field.

muhammad-ammar: These is not `test_models.py`. We do models related testing in `test_api.py`

edxval/models.py
```diff
@@ -199,12 +201,45 @@ def deconstruct(self):
...
+        return py_list
+
+
 class VideoImage(TimeStampedModel):
     """
     Image model for course video.
     """
     course_video = models.OneToOneField(CourseVideo, related_name="video_image")
     image = CustomizableImageField()
+    urls = ListField(max_length=2000)
```
muhammad-ammar: no it can not. our urls looks like `video-images/390326dfb3b24486b23f6dd1ed47e997.jpg`

muhammad-ammar: but we can add more validation on list if we want like below

```python
if len(py_list) > 3:
    raise ValidationError(u'list must not contain more than 2 items')

if reduce(lambda a, b: a + len(b), py_list, 0) > 2000:
    raise ValidationError(u'list items cumulative size exceeds 2000.')
```

Qubad786: I meant a standard url, and yes it'd be nice to limit the list to 3 so that `max_length=2000` can't be crossed.

muhammad-ammar:
@Qubad786 Added more validation in `ListField`

edxval/tests/test_api.py
Qubad786: I think there should be test that ensures our custom field works as expected when accessed/updated via related object – `course_video.video_image.urls`. As we've bitten by had such issues when using `CourseKeyField`.

muhammad-ammar: @Qubad786 Already testing this [at](https://github.com/edx/edx-val/pull/73/files#diff-630c04a9c9f1eb912f106ba7c548cc1eR1355).

edxval/models.py
```diff
@@ -199,12 +203,51 @@ def deconstruct(self):
...
+        return py_list
+
+
 class VideoImage(TimeStampedModel):
     """
     Image model for course video.
     """
     course_video = models.OneToOneField(CourseVideo, related_name="video_image")
     image = CustomizableImageField()
+    urls = ListField(max_length=LIST_ITEMS_LENGTH_SUM)
```
muzaffaryousaf: We need to choose proper name for this field which should be readable enough and explains the purpose.

muzaffaryousaf: I think we'll be storing names of the images rather than full CDN urls. What about `generated_images` and image to `active_image` or `selected_image` ? It's your choice.

muhammad-ammar:
@muzaffaryousaf @mushtaqak @Qubad786 Feedback addressed.

edxval/models.py
```diff
@@ -222,32 +221,51 @@ def to_python(self, value):
...
-            raise ValidationError(u'Must be a valid list of strings.')
+                self.validate(py_list)
+            except (ValueError, TypeError):
+                raise ValidationError(u'Must be a valid list of strings.')
 
         return py_list
 
+    def validate(self, value):
+        """
+        Validate data before saving to databse.
```
Qubad786: Typo: database

edxval/models.py
```diff
@@ -222,32 +221,51 @@ def to_python(self, value):
...
+
+        Arguemtns:
+            value(list): list to be validated
+
+        Returns:
+            list if validation is successfull
```
Qubad786: typo: successful

PR #79, opened 2017-07-11
Upgrade DRF version.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-1604

As part of an upgrade of django-rest-framework for edx-platform, this dependency needs to have its DRF version upgraded as well. Upgraded the DRF version, fixed tests, and bumped the version number.

### DIFF
--- a/edxval/serializers.py
+++ b/edxval/serializers.py
@@ -100,7 +100,7 @@ class CourseSerializer(serializers.RelatedField):
         Convert data into CourseVideo instance and image filename tuple.
         """
         course_id = data
-        course_video = image = None
+        course_video = image = ''
         if data:
             if isinstance(data, dict):
                 (course_id, image), = data.items()
--- a/edxval/tests/test_views.py
+++ b/edxval/tests/test_views.py
@@ -498,7 +498,7 @@ class VideoListTest(APIAuthTestCase):
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(
             response.data.get("edx_video_id")[0],
-            "This field must be unique."
+            u'video with this edx video id already exists.'
         )
         videos = len(self.client.get("/edxval/videos/").data)
         self.assertEqual(videos, 1)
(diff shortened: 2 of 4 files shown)


edxval/serializers.py
```diff
@@ -100,7 +100,7 @@ def to_internal_value(self, data):
         Convert data into CourseVideo instance and image filename tuple.
         """
         course_id = data
-        course_video = image = None
+        course_video = image = ''
```
muhammad-ammar: Any particular reason for empty string instead of `None`?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 django>=1.8,<1.9
-djangorestframework>=3.1,<3.2
+git+https://github.com/edx/django-rest-framework.git@1ceda7c086fddffd1c440cc86856441bbf0bd9cb#egg=djangorestframework==3.6.3
```
mushtaqak: Why are we using this drf commit? If there is any necessity then is there any plan to upgrade to stable release than one particular commit ?

edxval/tests/test_views.py
```diff
@@ -498,7 +498,7 @@ def test_post_videos(self):
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(
             response.data.get("edx_video_id")[0],
-            "This field must be unique."
+            u'video with this edx video id already exists.'
```
mushtaqak: @doctoryes why did we need to change this ?

edxval/serializers.py
doctoryes: The newest version of DRF (v3.6.3) is apparently stricter about empty strings vs. `None`. Defaulting to `None` and passing `None` through the DRF machinery caused a few tests to fail in edx-platform, while using an empty string fixed the tests.

requirements.txt
doctoryes: The reason why is documented here:
https://openedx.atlassian.net/browse/PLAT-1581?focusedCommentId=270321&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-270321
It'd be *great* to use the stable release again - perhaps we can get to that point soon.

edxval/tests/test_views.py
doctoryes: This test raises a DRF `ValidationError` and the message associated with that exception was changed to let the Django-generated message be returned. So I changed the test to check for the new more-specific message.

doctoryes: The code:
https://github.com/encode/django-rest-framework/blob/3.6.3/rest_framework/exceptions.py#L121-L140

PR #83, opened 2017-07-19
remove branches section, add env section

iloveagent57:
Get travis to build and deploy to PyPI on tag pushes.

@bmedx or @doctoryes would you mind giving this a review?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,16 @@
 language: python
 python:
 - '2.7'
+env:
+- TOXENV=django18
+- TOXENV=django110
+- TOXENV=django111
 install:
 - pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,15 +1,19 @@
 language: python
 python:
 - '2.7'
+env:
+- TOXENV=django18
+- TOXENV=django110
+- TOXENV=django111
+matrix:
+  allow_failures:
+  - env: TOXENV=django110
```
doctoryes: Is this necessary? In other words, do failures actually exist when testing against Django 1.10?

iloveagent57: ah, no, that's copypasta from edx-submission

PR #89, opened 2017-08-03
Add course-wide Transcript Preference

Qubad786:
## [EDUCATOR-1092](https://openedx.atlassian.net/browse/EDUCATOR-1092) 

1. Add course-wide Transcript Preference
2. Prepare 3rd party transcription plans
3. Api methods to use transcript prefs/plans in edx-platform
4. Add an endpoint to update video status (for VEDA)
5. Refactor ListField to support `max_items`


Qubad786:
@muzaffaryousaf

edxval/views.py
```diff
@@ -162,6 +170,52 @@ def post(self, request):
         return response
 
 
+class VideoStatusView(APIView):
+    """
+    A Video View, used by VEDA to update the status of a video.
+
+    Note:
+        Currently, the valid statuses are `transcription_in_progress` and `transcription_ready` because it was intended
+        to only be used for video transcription process by VEDA but if you found it helpful to your needs, you can add
```
muzaffaryousaf: VEDA --> edx-video-pipeline

edxval/views.py
```diff
@@ -13,13 +15,19 @@
...
     VideoSerializer,
     TranscriptSerializer,
 )
 
+VALID_VIDEO_STATUSES = [
+    'transcription_in_progress',
+    'transcription_ready',
+]
+
+logger = logging.getLogger(__name__)  # pylint: disable=C0103
```
muzaffaryousaf: Never being used. Do we need to ?

edxval/api.py
```diff
@@ -230,6 +232,54 @@ def update_video_transcript(video_id, language, transcript_url, transcript_forma
...
+
+def get_transcript_preferences(course_id):
+    """
+    Retrieves course wide transcript preferences
+
+    Arguments:
+        course_id (CourseKey): a course key
+    """
+    transcript_preference = None
+    transcript_preference_set = TranscriptPreference.objects.filter(course_id=course_id)
```
muzaffaryousaf: Why filter ? I think .get would be sufficient Or this is just to avoid exception?

edxval/api.py
```diff
@@ -230,6 +232,54 @@ def update_video_transcript(video_id, language, transcript_url, transcript_forma
...
+def create_or_update_transcript_preferences(course_id, provider, **kwargs):
+    """
+    Creates or updates course-wide transcript preferences
+
+    Arguments:
+        course_id(CourseKey): a course key
+        provider(str): transcript provider 3play/cielo24
+
+    Keyword Arguments:
+        kwargs(dict): keyword arguments
```
muzaffaryousaf: `keyword arguments` --> what are you expecting in this kwarg would be better.

edxval/api.py
```diff
@@ -230,6 +232,54 @@ def update_video_transcript(video_id, language, transcript_url, transcript_forma
...
+    """
+    try:
+        transcript_preference = TranscriptPreference.objects.get(course_id=course_id)
+        transcript_preference.provider = provider
+        transcript_preference.three_play_turnaround = kwargs.get('three_play_turnaround')
+        transcript_preference.cielo24_turnaround = kwargs.get('cielo24_turnaround')
+        transcript_preference.cielo24_fidelity = kwargs.get('cielo24_fidelity')
+        transcript_preference.preferred_languages = kwargs.get('preferred_languages')
```
muzaffaryousaf: Is it possible to have None or empty value in these kwargs, if yes then how about default values.

edxval/views.py
```diff
@@ -162,6 +170,52 @@ def post(self, request):
...
+            response_payload = {}
+        except Video.DoesNotExist:
+            response_status = status.HTTP_400_BAD_REQUEST
+            response_payload = {
+                'message': u'Video is not found for specified edx_video_id: {edx_video_id}'.format(
+                    edx_video_id=edx_video_id
+                )
+            }
+
+        return Response(status=response_status, data=response_payload)
```
muzaffaryousaf: Let's make it 200 and have another response in except block to return 400 and respective payload, since we're already doing it on many places.

edxval/serializers.py
```diff
@@ -193,3 +193,20 @@ def update(self, instance, validated_data):
...
         return instance
+
+
+class TranscriptPreferenceSerializer(serializers.ModelSerializer):
+    """
+    Serializer for Transcript Preference object
+    """
+    class Meta:  # pylint: disable=C1001, C0111
+        model = TranscriptPreference
+        fields = (
```
mushtaqak: We should remove `fields` option as this would give error on serializer : `AssertionError: Cannot set both 'fields' and 'exclude' options on serializer TranscriptPreferenceSerializer.`

edxval/api.py
```diff
@@ -143,6 +151,135 @@ def update_video_status(edx_video_id, status):
...
+    """
+    transcript_preference = None
+    transcript_preference_set = TranscriptPreference.objects.filter(course_id=course_id)
+    if transcript_preference_set.exists():
+        transcript_preference = TranscriptPreferenceSerializer(transcript_preference_set.first()).data
+
+    return transcript_preference
+
+
+def create_or_update_transcript_preferences(course_id, provider, **kwargs):
```
muzaffaryousaf: kwargs is dict of preferences, so let's make it preferences.

edxval/api.py
```diff
@@ -143,6 +151,135 @@ def update_video_status(edx_video_id, status):
...
+
+    Arguments:
+        course_id(CourseKey): a course key
+        provider(str): transcript provider 3play/cielo24
+
+    Keyword Arguments:
+        kwargs(dict): keyword arguments
+    """
+    try:
+        transcript_preference = TranscriptPreference.objects.get(course_id=course_id)
```
muzaffaryousaf: let's use get_or_create.

edxval/api.py
```diff
@@ -226,6 +228,54 @@ def create_or_update_video_transcript(
...
+        transcript_preference.three_play_turnaround = kwargs.get('three_play_turnaround')
+        transcript_preference.cielo24_turnaround = kwargs.get('cielo24_turnaround')
+        transcript_preference.cielo24_fidelity = kwargs.get('cielo24_fidelity')
+        transcript_preference.preferred_languages = kwargs.get('preferred_languages')
+    except TranscriptPreference.DoesNotExist:
+        transcript_preference = TranscriptPreference(course_id=course_id, provider=provider, **kwargs)
+
+    transcript_preference.save()
+
+    return TranscriptPreferenceSerializer(transcript_preference).data
```
muzaffaryousaf: What if there is an error during save ? Should we raise an exception to capture it in platform.

muhammad-ammar:
@muzaffaryousaf @Qubad786 This is ready for review.

edxval/views.py
```diff
@@ -159,6 +165,52 @@ def post(self, request):
         return response
 
 
+class VideoStatusView(APIView):
```
muzaffaryousaf: Don't we need decorator / permissions class for this view ?

edxval/serializers.py
```diff
@@ -4,10 +4,12 @@
 Serialization is usually sent through the VideoSerializer which uses the
 EncodedVideoSerializer which uses the profile_name as it's profile field.
 """
+import json
```
muzaffaryousaf: can we remove this un-used import ?

edxval/tests/test_views.py
```diff
@@ -990,3 +990,48 @@ def test_error_responses(self, post_data, message):
...
+            'status_code': status.HTTP_400_BAD_REQUEST,
+        },
+        {
+            'patch_data': {'edx_video_id': 'super-soaker', 'status': 'transcription_ready'},
+            'message': None,
+            'status_code': status.HTTP_200_OK,
+        },
+    )
+    @unpack
+    def test_create_transcript(self, patch_data, message, status_code):
```
muzaffaryousaf: test_transcript_status ?

PR #98, opened 2017-09-19
remove external kwarg for import

muhammad-ammar:


edxval/tests/test_api.py
```diff
@@ -1304,18 +1304,14 @@ def test_external_video_transcript(self):
...
     def test_external_no_video_transcript(self):
         """
         Verify that transcript import for external video working as expected when there is no transcript.
         """
-        api.import_from_xml(
-            etree.fromstring('<video_asset/>'),
-            'does_not_exist_id',
-            external=True
-        )
+        api.import_from_xml(etree.fromstring('<video_asset/>'), '')
```
muzaffaryousaf: do we need this test when we don't have external ? If yes then we need to change the test name.

muhammad-ammar: This test is only needed for external videos.

PR #99, opened 2017-09-20
Organization specific transcript credentials state

muhammad-ammar:
[EDUCATOR-1376](https://openedx.atlassian.net/browse/EDUCATOR-1376)


edxval/api.py
```diff
@@ -143,6 +143,55 @@ def update_video_status(edx_video_id, status):
...
+    if credentials_state.exists():
+        credentials_state_info = {}
+        for credential_state in credentials_state:
+            credentials_state_info['org'] = credential_state.org
+            credentials_state_info[credential_state.provider] = credential_state.exists
+
+        return credentials_state_info
+
+
+def update_organization_transcript_credentials_state(org, provider, exists):
```
mushtaqak: I would rename it to `update_transcript_organization_credentials_state`

edxval/api.py
```diff
@@ -143,6 +143,55 @@ def update_video_status(edx_video_id, status):
...
+    """
+    Update transcript credentials state for a course organization.
```
mushtaqak: Typo: Updates

edxval/models.py
```diff
@@ -605,6 +605,48 @@ def __unicode__(self):
...
 
 
+class OrganizationTranscriptCredentialsState(models.Model):
+    """
+    State of transcript credentials for a course organization
+    """
+    class Meta:
+        unique_together = ('org', 'provider')
+
+    org = models.CharField(verbose_name='Course Organization', max_length=255)
```
mushtaqak: `max_length=255` isn't that too much

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,87 @@ def test_create_transcript_preferences(self):
...
+        """
+        state_info = api.get_organization_transcript_credentials_state(org=org, provider=provider)
+        credentials_state = OrganizationTranscriptCredentialsState.objects.get(org=org, provider=provider)
+        self.assertEqual(state_info['org'], credentials_state.org)
+        self.assertEqual(state_info[provider], credentials_state.exists)
+
+    def test_get_credentials_state_with_org_only(self):
+        """
+        Verify that `get_organization_transcript_credentials_state`
+        method works as expected when pnly org is given
```
mushtaqak: `pnly` --> `only`

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,87 @@ def test_create_transcript_preferences(self):
...
+        {'org': 'edX', 'provider': 'Cielo24'},
+        {'org': 'edX', 'provider': '3PlayMedia'},
+    )
+    @unpack
+    def test_get_credentials_state(self, org, provider):
+        """
+        Verify that `get_organization_transcript_credentials_state` method works as expected
+        """
+        state_info = api.get_organization_transcript_credentials_state(org=org, provider=provider)
+        credentials_state = OrganizationTranscriptCredentialsState.objects.get(org=org, provider=provider)
```
mushtaqak: We may not need to query db, if we add a `exists` parameter in ddt. Something like :
```
@data(
         {'org': 'edX', 'provider': 'Cielo24', 'exists': False},
         {'org': 'edX', 'provider': '3PlayMedia', 'exists': False},
     )
     @unpack
     def test_get_credentials_state(self, org, provider, exists):
```

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,87 @@ def test_create_transcript_preferences(self):
...
+        self.assertEqual(state_info['org'], credentials_state.org)
+        self.assertEqual(state_info[provider], credentials_state.exists)
+
+    def test_get_credentials_state_with_org_only(self):
```
mushtaqak: This test can be combined in above test with proper ddt

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,87 @@ def test_create_transcript_preferences(self):
...
+        credentials_state = OrganizationTranscriptCredentialsState.objects.get(org=kwargs['org'])
+        for key in kwargs:
+            self.assertEqual(getattr(credentials_state, key), kwargs[key])
+
+    @data(
+        {'org': 'edX', 'provider': 'Cielo24'},
+        {'org': 'edX', 'provider': '3PlayMedia'},
+    )
+    @unpack
+    def test_get_credentials_state(self, org, provider):
```
mushtaqak: We should also add a test if org does not match, what is expected result we should expect

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,87 @@ def test_create_transcript_preferences(self):
...
+        Tests setup
+        """
+        OrganizationTranscriptCredentialsState.objects.create(
+            org='edX', provider='Cielo24', exists=False
+        )
+        OrganizationTranscriptCredentialsState.objects.create(
+            org='edX', provider='3PlayMedia', exists=False
+        )
+
+    def assert_credential_state():
```
mushtaqak: `assert_credential_state` is not used anywhere

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,87 @@ def test_create_transcript_preferences(self):
...
+            OrganizationTranscriptCredentialsState.objects.create(
+                org='edX', provider=provider, exists=False
+            )
+
+    @data(
+        {'org': 'MAX', 'provider': 'Cielo24', 'exists': True},
+        {'org': 'MAX', 'provider': '3PlayMedia', 'exists': True},
+    )
+    @unpack
+    def test_credentials_state_update(self, **kwargs):
```
mushtaqak: We should also add a test the expected behaviour when org does not match

edxval/models.py
```diff
@@ -605,6 +605,48 @@ def __unicode__(self):
         return u'{course_id} - {provider}'.format(course_id=self.course_id, provider=self.provider)
 
 
+class OrganizationTranscriptCredentialsState(models.Model):
```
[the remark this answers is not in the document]
muhammad-ammar: These are transcription credentials for an organization, So that's why I used the name `OrganizationTranscriptCredentialsState`

edxval/models.py
```diff
@@ -605,6 +605,48 @@ def __unicode__(self):
...
+        return instance, created
+
+    def __unicode__(self):
+        return u'OrganizationTranscriptCredentialsState(org={}, provider={}, exists={})'.format(
+            self.org, self.provider, self.exists
+        )
+
+
+@receiver(models.signals.pre_save, sender=OrganizationTranscriptCredentialsState)
+def validate_provider(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
muhammad-ammar: This is the cleanest way I found to validate `provider` value before saving.

edxval/tests/test_api.py
muhammad-ammar: We need reading from db to verify that data received from api method is same as stored in db.

edxval/tests/test_api.py
muhammad-ammar: In case of update/create, new credentials state record will be created if org does not match.

muhammad-ammar:
@mushtaqak Feedback Addressed

edxval/models.py
```diff
@@ -633,15 +633,22 @@ def update_or_create(cls, org, provider, exists):
         return instance, created
 
     def __unicode__(self):
-        return u'OrganizationTranscriptCredentialsState(org={}, provider={}, exists={})'.format(
-            self.org, self.provider, self.exists
+        """
+        Returns unicode representation of provider credentials state for an organization.
+
+        NOTE: Message will look like below:
```
mushtaqak: I am not sure if this the best way ?

edxval/tests/test_api.py
mushtaqak: I think we missed this test for `No org found` case?

edxval/models.py
```diff
@@ -605,6 +605,55 @@ def __unicode__(self):
...
+            defaults={'exists': exists},
+        )
+
+        return instance, created
+
+    def __unicode__(self):
+        """
+        Returns unicode representation of provider credentials state for an organization.
+
+        NOTE: Message will look like below:
```
mushtaqak: I am not sure if this is the best way to show record name ?

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,93 @@ def test_create_transcript_preferences(self):
...
+                org='edX', provider=provider, exists=False
+            )
+
+    @data(
+        {'org': 'MAX', 'provider': 'Cielo24', 'exists': True},
+        {'org': 'MAX', 'provider': '3PlayMedia', 'exists': True},
+        {'org': 'edx', 'provider': '3PlayMedia', 'exists': True},
+    )
+    @unpack
+    def test_credentials_state_update(self, **kwargs):
```
muhammad-ammar: @mushtaqak update this test for an existing org.

edxval/models.py
```diff
@@ -605,6 +605,55 @@ def __unicode__(self):
         return u'{course_id} - {provider}'.format(course_id=self.course_id, provider=self.provider)
 
 
+class OrganizationTranscriptCredentialsState(models.Model):
```
mushtaqak: Also, we should probably use `TimeStampedModel` to keep track of created and modified time

edxval/models.py
```diff
@@ -605,6 +605,55 @@ def __unicode__(self):
...
+    """
+    State of transcript credentials for a course organization
+    """
+    class Meta:
+        unique_together = ('org', 'provider')
+
+    org = models.CharField(verbose_name='Course Organization', max_length=32)
+    provider = models.CharField(
```
Qubad786: @muhammad-ammar we don't need to verify provider on pre-save signal. Model will automatically raises Exception if the you try to save it with a Provider which is not in its choices.

edxval/models.py
```diff
@@ -605,6 +605,55 @@ def __unicode__(self):
...
+        NOTE: Message will look like below:
+            edX has Cielo24 credentials
+            edX has no 3PlayMedia credentials
+        """
+        return u'{org} {state} {provider} credentials'.format(
+            org=self.org, provider=self.provider, state='has' if self.exists else 'has no'
+        )
+
+
+@receiver(models.signals.pre_save, sender=OrganizationTranscriptCredentialsState)
```
Qubad786: No need for this.

edxval/models.py
```diff
@@ -605,6 +605,55 @@ def __unicode__(self):
...
+    def __unicode__(self):
+        """
+        Returns unicode representation of provider credentials state for an organization.
+
+        NOTE: Message will look like below:
+            edX has Cielo24 credentials
+            edX has no 3PlayMedia credentials
+        """
+        return u'{org} {state} {provider} credentials'.format(
+            org=self.org, provider=self.provider, state='has' if self.exists else 'has no'
```
Qubad786: `doesn't have` instead of `has no` sounds better.

edxval/admin.py
```diff
@@ -48,9 +48,16 @@ class CourseVideoAdmin(admin.ModelAdmin):
     verbose_name_plural = 'Course Videos'
 
 
+class OrganizationTranscriptCredentialsStateAdmin(admin.ModelAdmin):
+    model = OrganizationTranscriptCredentialsState
+    verbose_name = 'Organization Transcript Credential State'
+    verbose_name_plural = 'Organization Transcript Credentials State'
```
Qubad786: I think we need to pluralize `State` instead of `Credentials` -- L53 & L54.

edxval/api.py
```diff
@@ -143,6 +143,55 @@ def update_video_status(edx_video_id, status):
     video.save()
 
 
+def get_organization_transcript_credentials_state(org, provider=None):
```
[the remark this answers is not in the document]
Qubad786: `get_3rd_party_transcript_credentials_state_for_org` would be good.

edxval/api.py
```diff
@@ -143,6 +143,55 @@ def update_video_status(edx_video_id, status):
...
+    """
+    Returns transcript credentials state for a course organization.
+
+    Arguments:
+        org (unicode): course organization
+        provider (unicode): transcript provider
+
+    Returns:
+        dict: contains the information regarding organization's transcript provider credentials
```
Qubad786: organization's -->  organization-specific

edxval/api.py
```diff
@@ -143,6 +143,55 @@ def update_video_status(edx_video_id, status):
...
+    Returned dict will look like one of below
+    {
+        'org': 'edX',
+        'Cielo24': True,
+    }
+
+    {
+        'org': 'edX',
+        'Cielo24': False,
+        '3PlayMedia': True,
```
Qubad786: @muhammad-ammar this is also an idea, can we just change our model like this?
```
org
has_3play_creds
has_cielo_creds
```

edxval/api.py
```diff
@@ -143,6 +143,55 @@ def update_video_status(edx_video_id, status):
...
+def update_organization_transcript_credentials_state(org, provider, exists):
+    """
+    Updates transcript credentials state for a course organization.
+
+    Arguments:
+        org (unicode): course organization
+        provider (unicode): transcript provider
+        exists (bool): state of credentials
+    """
+    OrganizationTranscriptCredentialsState.update_or_create(org, provider, exists)
```
Qubad786: return something? like updated or created record?

edxval/api.py
Qubad786: `update_3rd_party_transcript_credentials_state_for_org`

edxval/tests/test_api.py
```diff
@@ -2035,3 +2035,93 @@ def test_create_transcript_preferences(self):
...
+        """
+        Verify that `update_organization_transcript_credentials_state` method works as expected
+        """
+        api.update_organization_transcript_credentials_state(
+            org=kwargs['org'], provider=kwargs['provider'], exists=kwargs['exists']
```
Qubad786: simpler: `api.update_organization_transcript_credentials_state(**kwargs)`

edxval/models.py
```diff
@@ -605,6 +605,55 @@ def __unicode__(self):
...
 
 
+class OrganizationTranscriptCredentialsState(models.Model):
+    """
+    State of transcript credentials for a course organization
+    """
+    class Meta:
+        unique_together = ('org', 'provider')
+
+    org = models.CharField(verbose_name='Course Organization', max_length=32)
```
mushtaqak: `max_length=50` to be consistent with edx-pipeline

mushtaqak: This has been done in https://github.com/edx/edx-val/pull/102

edxval/migrations/0008_auto_20170922_0238.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
mushtaqak: This would also need to change as 0006 migration has been renamed in https://github.com/edx/edx-val/pull/102

edxval/api.py
Qubad786: no need, looks good.

edxval/models.py
muhammad-ammar: This is not happening. Model allows to save the incorrect data if I remove the signal.

muhammad-ammar:
@muzaffaryousaf @Qubad786 @mushtaqak Feedback addressed. Whoever is using this needs to update the hash and api function names.

muzaffaryousaf:
@Qubad786 @mushtaqak can you guys finalize this PR as per the usage in our other branches.

muzaffaryousaf:
@mushtaqak @Qubad786 This needs review from you guys.

edxval/models.py
```diff
@@ -612,6 +612,55 @@ def __unicode__(self):
...
+
+        return instance, created
+
+    def __unicode__(self):
+        """
+        Returns unicode representation of provider credentials state for an organization.
+
+        NOTE: Message will look like below:
+            edX has Cielo24 credentials
+            edX has no 3PlayMedia credentials
```
Qubad786: nit: `edX doesn't have 3PlayMedia credentials`.

edxval/models.py
Qubad786: @muhammad-ammar that's may be because of sqlite db.

edxval/models.py
```diff
@@ -612,6 +612,55 @@ def __unicode__(self):
...
+    )
+    exists = models.BooleanField(default=False, help_text='Transcript credentials state')
+
+    @classmethod
+    def update_or_create(cls, org, provider, exists):
+        """
+        Update or create credentials state.
+        """
+        instance, created = cls.objects.update_or_create(
+            org=org, provider=provider,
```
Qubad786: nit: proper line break for `kwargs`.

edxval/models.py
```diff
@@ -612,6 +612,55 @@ def __unicode__(self):
...
+        )
+
+
+@receiver(models.signals.pre_save, sender=ThirdPartyTranscriptCredentialsState)
+def validate_provider(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Validate that input provider value must be one of TranscriptProviderType.CHOICES.
+    """
+    if instance.provider not in dict(TranscriptProviderType.CHOICES).keys():
+        raise ValidationError('Invalid transcript provider value')
```
Qubad786: @muhammad-ammar I think its alright to skip the this signal to validate the data. We can validate it in api utils which are responsible for creating an actual state record in db, Or even we may not validate as its getting validated from the view which is calling val's api.

edxval/migrations/0007_auto_20171005_0154.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
mushtaqak: I think we should rename the migration to better name ? just as we did for `006_`

edxval/models.py
```diff
@@ -612,6 +612,55 @@ def __unicode__(self):
         return u'{course_id} - {provider}'.format(course_id=self.course_id, provider=self.provider)
 
 
+class ThirdPartyTranscriptCredentialsState(models.Model):
```
mushtaqak: We should use `TimeStampedModel`

muhammad-ammar:
@mushtaqak @Qubad786 Feedback addressed.

muhammad-ammar:
@mushtaqak @Qubad786 Feedback addressed.

edxval/api.py
```diff
@@ -143,6 +143,53 @@ def update_video_status(edx_video_id, status):
...
+    query_filter = {'org': org}
+    if provider:
+        query_filter['provider'] = provider
+
+    try:
+        credentials_state = {
+            credential.provider: credential.exists
+            for credential in ThirdPartyTranscriptCredentialsState.objects.filter(**query_filter)
+        }
+    except ThirdPartyTranscriptCredentialsState.DoesNotExist:
```
Qubad786: filter won't raise `ThirdPartyTranscriptCredentialsState.DoesNotExist`. We are good without `try: except..` here.

I think just return the following would work.
```bash
          {
              credential.provider: credential.exists
              for credential in ThirdPartyTranscriptCredentialsState.objects.filter(**query_filter)
         }
```

PR #100, opened 2017-09-22
add video_source_language field

muhammad-ammar:


muhammad-ammar:
@mushtaqak @Qubad786 @muzaffaryousaf This is ready for review.

edxval/models.py
```diff
@@ -600,6 +600,13 @@ class TranscriptPreference(TimeStampedModel):
         blank=True,
     )
     preferred_languages = ListField(verbose_name='Preferred Languages', max_items=50, default=[], blank=True)
+    video_source_language = models.CharField(
+        verbose_name='Video Source Language',
+        max_length=20,
+        null=True,
+        blank=True,
+        help_text='This is the source language of a video.'
```
muzaffaryousaf: can we also add that this for 3play for now.

Qubad786: I agree, that'd be great.

edxval/models.py
```diff
@@ -600,6 +600,13 @@ class TranscriptPreference(TimeStampedModel):
         blank=True,
     )
     preferred_languages = ListField(verbose_name='Preferred Languages', max_items=50, default=[], blank=True)
+    video_source_language = models.CharField(
+        verbose_name='Video Source Language',
+        max_length=20,
```
Qubad786: we should follow the same convention for language codes -- I think [this](https://github.com/edx/edx-val/pull/100/files#diff-bf00c7cdc19274a54f59d0526f065c8dR409) is what we decided in the first place. cc @mushtaqak

PR #102, opened 2017-10-02
Change max_length for language_code for transcript code.

muzaffaryousaf:
Should have single migration for this feature.

EDU-1415


muzaffaryousaf:
@Qubad786 @muhammad-ammar  Quick look ?

PR #103, opened 2017-10-09
help text for source language.

Qubad786:
[EDUCATOR-1491](https://openedx.atlassian.net/browse/EDUCATOR-1491)

The `video_source_language` also accounts for Cielo24. So, making the help text more generic.


Qubad786:
@muzaffaryousaf have a look.

PR #104, opened 2017-10-12
VAL admin optimizations

muzaffaryousaf:
Update list_display for Course Video, Course Image, Video Transcript, Transcript Preferences.
Use Raw display id widget for course_video inside Video Image Admin model to optimize.

EDU-1358

### DIFF
--- a/edxval/admin.py
+++ b/edxval/admin.py
@@ -1,7 +1,7 @@
 """
 Admin file for django app edxval.
 """
-
+from django import forms
 from django.contrib import admin
 
 from .models import (CourseVideo, EncodedVideo, Profile, TranscriptPreference,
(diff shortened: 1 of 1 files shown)


edxval/admin.py
```diff
@@ -36,13 +36,40 @@ class VideoAdmin(admin.ModelAdmin):  # pylint: disable=C0111
...
+    class Meta:
+        model = VideoImage
+        fields = ('course_video', 'image', 'generated_images')
+
+
 class VideoImageAdmin(admin.ModelAdmin):
+    list_display = ('get_course_id', 'image', 'generated_images')
+
+    def get_course_id(self, obj):
+        return obj.course_video.course_id
```
Qubad786: Shouldn't there be both course and video IDs?

Qubad786: something like `course_video`

muzaffaryousaf: We need to show the course id why course_video ?

PR #106, opened 2017-10-26
Remove subtitles code from VAL.

muzaffaryousaf:
EDU-1369


muzaffaryousaf:
Related app-permission PR https://github.com/edx/app-permissions/pull/497

PR #110, opened 2017-11-20
PLAT-1413 Remove obsolete TEMPLATE_* settings

jmbowman:
These settings are deprecated in Django 1.8 and removed in 1.10.  They were auto-generated by an old version of Django, and aren't even used in the tests which this file is meant for.


jmbowman:
@doctoryes Correct, I think they were just auto-generated by old versions of the `startproject` management command.

PR #113, opened 2017-12-04
api method for video transcript deletion

muhammad-ammar:
[EDUCATOR-2034]( https://openedx.atlassian.net/browse/EDUCATOR-2034)

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -336,6 +336,24 @@ def create_or_update_video_transcript(
     return video_transcript.url()
 
 
+def delete_video_transcript(video_id, language_code):
+    """
+    Delete transcript for an existing video.
+
+    Arguments:
+        video_id: id of the video with which transcript is associated
+        language_code: language code of a video transcript
+    """
+    try:
+        video_transcript = VideoTranscript.objects.get(video_id=video_id, language_code=language_code)
+        # delete the actual transcript file from storage
+        video_transcript.transcript.delete()
+        # delete the record from db
+        video_transcript.delete()
+    except VideoTranscript.DoesNotExist:
+        pass
+
+
 def get_3rd_party_transcription_plans():
     """
     Retrieves 3rd party transcription plans.
--- a/edxval/tests/test_api.py
+++ b/edxval/tests/test_api.py
@@ -3,9 +3,12 @@
 Tests for the API for Video Abstraction Layer
 """
 import json
+import os
+import tempfile
 
 import mock
 from ddt import data, ddt, unpack
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.files import File
 from django.core.files.base import ContentFile
@@ -16,18 +19,19 @@ from django.test import TestCase
 from lxml import etree
 from mock import patch
 from rest_framework import status
-from django.conf import settings
 
 from edxval import api as api
+from edxval import utils
 from edxval.api import (InvalidTranscriptFormat, InvalidTranscriptProvider,
                         SortDirection, ValCannotCreateError,
                         ValCannotUpdateError, ValVideoNotFoundError,
                         VideoSortField)
 from edxval.models import (LIST_MAX_ITEMS, CourseVideo, EncodedVideo, Profile,
-                           TranscriptFormat, TranscriptProviderType, Video,
-                           VideoImage, VideoTranscript, TranscriptPreference, ThirdPartyTranscriptCredentialsState)
+                           ThirdPartyTranscriptCredentialsState,
+                           TranscriptFormat, TranscriptPreference,
+                           TranscriptProviderType, Video, VideoImage,
+                           VideoTranscript)
 from edxval.tests import APIAuthTestCase, constants
-from edxval import utils
 
 
 FILE_DATA = """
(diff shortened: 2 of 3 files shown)


muhammad-ammar:
@iloveagent57 Any idea how we can fix this [build error](https://travis-ci.org/edx/edx-val/jobs/311247002)? 

```bash
Collecting Django>=1.4.2 (from django-model-utils==2.3.1->-r requirements.txt (line 5))
  Downloading Django-2.0.tar.gz (8.0MB)
    100% |████████████████████████████████| 8.0MB 185kB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-MtINs3/Django/setup.py", line 32, in <module>
        version = __import__('django').get_version()
      File "django/__init__.py", line 1, in <module>
        from django.utils.version import get_version
      File "django/utils/version.py", line 61, in <module>
        @functools.lru_cache()
    AttributeError: 'module' object has no attribute 'lru_cache'
``` 

As per the above logs, pip is fetching `Django-2.0` which has no support for python2.7. I think this issue can be resolved if we somehow configure the build to install edxval supported Django(`Django>=1.8,<2`) first and then install the other requirements, But I don't know how to do that?

muhammad-ammar:
@edx/educator-mallow This is ready for review.

edxval/api.py
```diff
@@ -336,6 +336,17 @@ def create_or_update_video_transcript(
     return video_transcript.url()
 
 
+def delete_video_transcript(video_id, language_code):
+    """
+    Delete transcript for an existing video.
+
+    Arguments:
+        video_id: it can be an edx_video_id or an external_id extracted from external sources in a video component.
```
Qubad786: need to change this comment.

edxval/api.py
```diff
@@ -336,6 +336,17 @@ def create_or_update_video_transcript(
...
+        language_code: language code of a video transcript
+    """
+    VideoTranscript.objects.filter(video_id=video_id, language_code=language_code).delete()
```
Qubad786: Does this delete actual content from S3?

edxval/tests/test_api.py
```diff
@@ -1954,6 +1954,18 @@ def test_get_available_transcript_languages(self):
         transcript_languages = api.get_available_transcript_languages(video_ids=video_ids)
         self.assertItemsEqual(transcript_languages, ['de', 'en', 'ur'])
 
+    def test_delete_video_transcript(self):
```
Qubad786: We should also assert the transcript content.

edxval/tests/test_api.py
```diff
@@ -1954,6 +1954,18 @@ def test_get_available_transcript_languages(self):
...
+        """
+        Verify that `delete_video_transcript` works as expected.
+        """
+        query_filter = {
+            'video_id': 'super-soaker',
+            'language_code': 'de'
+        }
+        self.assertEqual(VideoTranscript.objects.filter(**query_filter).count(), 1)
+        api.delete_video_transcript(**query_filter)
+        self.assertEqual(VideoTranscript.objects.filter(**query_filter).count(), 0)
```
Qubad786: Should we do the following?
```bash
with self.assertRaises(VideoTranscript.DoesNotExist):
     VideoTranscript.objects.get(**query_filter)
```

muhammad-ammar:
@Qubad786 Addressed your feedback.

edxval/api.py
```diff
@@ -336,6 +336,24 @@ def create_or_update_video_transcript(
...
+def delete_video_transcript(video_id, language_code):
+    """
+    Delete transcript for an existing video.
+
+    Arguments:
+        video_id: id of the video with which transcript is associated
+        language_code: language code of a video transcript
+    """
+    try:
+        video_transcript = VideoTranscript.objects.get(video_id=video_id, language_code=language_code)
```
mushtaqak: What if video id and language code is not provided? should we not return some error so that frontend can let user know ?

edxval/api.py
```diff
@@ -336,6 +336,24 @@ def create_or_update_video_transcript(
     return video_transcript.url()
 
 
+def delete_video_transcript(video_id, language_code):
```
mushtaqak: Should this not notify if transcript was deleted or some error message ?

edxval/tests/test_api.py
```diff
@@ -1952,7 +1980,21 @@ def test_get_available_transcript_languages(self):
         # `non_existent_video_id` that does not have transcript
         video_ids = ['super-soaker', self.video_id, dupe_lang_video_id, 'non_existent_video_id']
         transcript_languages = api.get_available_transcript_languages(video_ids=video_ids)
-        self.assertItemsEqual(transcript_languages, ['de', 'en', 'ur'])
+        self.assertItemsEqual(transcript_languages, ['de', 'en', 'ur', 'fr'])
+
+    def test_delete_video_transcript(self):
```
mushtaqak: We should also add some test if no video id and language is passed to see this does not crash.

edxval/api.py
mushtaqak: As discussed, we will notify and handle error scenarios in platform side.

PR #118, opened 2018-01-26
improve video transcripts admin

Qubad786:
**Making transcript searchable and editable –– fixes hanging on the video dropdown.**
_Screenshots:_
screen shot 2018-01-26 at 4 26 04 pm
screen shot 2018-01-26 at 4 26 38 pm
screen shot 2018-01-26 at 4 26 47 pm
screen shot 2018-01-26 at 4 27 06 pm

### DIFF
--- a/edxval/admin.py
+++ b/edxval/admin.py
@@ -69,9 +69,19 @@ class CourseVideoAdmin(admin.ModelAdmin):
 
 
 class VideoTranscriptAdmin(admin.ModelAdmin):
-    list_display = ('video', 'language_code', 'provider', 'file_format')
+    raw_id_fields = ('video',)
+    list_display = ('get_video', 'language_code', 'provider', 'file_format')
+    search_fields = ('id', 'video__edx_video_id', 'language_code')
+
+    def get_video(self, transcript):
+        return transcript.video.edx_video_id if getattr(transcript, 'video', False) else ''
+
+    get_video.admin_order_field = 'video'
+    get_video.short_description = 'Video'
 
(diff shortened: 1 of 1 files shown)


Qubad786:
@irfanuddinahmad @muhammad-ammar please have a look.

edxval/admin.py
```diff
@@ -69,9 +69,19 @@ def get_video_id(self, obj):
 
 
 class VideoTranscriptAdmin(admin.ModelAdmin):
-    list_display = ('video', 'language_code', 'provider', 'file_format')
+    raw_id_fields = ('video',)
+    list_display = ('get_video', 'language_code', 'provider', 'file_format')
+    search_fields = ('id', 'video__edx_video_id', 'language_code')
```
mushtaqak: Should we add `id` field to search ?

edxval/admin.py
```diff
@@ -69,9 +69,19 @@ def get_video_id(self, obj):
...
+
+    def get_video(self, transcript):
+        return transcript.video.edx_video_id if getattr(transcript, 'video', False) else ''
```
mushtaqak: This if-else seems unnecessary, I think we would always get an associated video with transcript. Isn't this the case ?

Qubad786: we have `null=True` constraint on video foreign key at the moment so that Django don't complain to have a default for existing rows. That's why did this.

edxval/admin.py
Qubad786: its alright in case if someone just has an ID.

PR #123, opened 2018-02-02
add missing audio_mp3 profile

muhammad-ammar:


edxval/migrations/0011_data__add_audio_mp3_profile.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations, models
+
+
+AUDIO_MP3_PROFILE = 'audio_mp3'
+
+
+def create_audio_mp3_profile(apps, schema_editor):
+    """ Create audio_mp3 profile """
+    Profile = apps.get_model("edxval", "Profile")
```
mushtaqak: nit: single quotes

edxval/tests/test_api.py
```diff
@@ -218,7 +218,7 @@ def test_create_profile(self):
         api.create_profile(constants.PROFILE_DESKTOP)
         profiles = list(Profile.objects.all())
         profile_names = [unicode(profile) for profile in profiles]
-        self.assertEqual(len(profiles), 7)
+        self.assertEqual(len(profiles), 8)
```
mushtaqak: I see that total [default profiles](https://github.com/edx/edx-val/blob/b62c3077439151670168b866bbb91a3f6660f591/edxval/migrations/0002_data__default_profiles.py#L7) are 5 in data migration and then HLS profile was added later, so there would be only 6 prior to adding mp3 profile?

muhammad-ammar: there are total of 7 profiles. In this test we are creating an additional profile and then verify that correct number of profiles exists.

muhammad-ammar:
@mushtaqak Feedback addressed

PR #131, opened 2018-03-01
PLAT-1985 Better dependency management

jmbowman:
Generate the pinned requirements files via pip-tools, and loosen the version constraints in setup.py to allow more flexibility in upgrading dependencies downstream.  Added a few other things while copying things over from the Django app cookiecutter:

* There is now a `quality` tox environment for running pylint, etc. (the checks don't pass yet)
* Switched the README to reStructuredText so it will be formatted properly on PyPI
* Upgraded all dependencies (which now won't force downstream code to upgrade them also)


jmbowman:
@mushtaqak I noticed that you're also doing some work in this repo, with an open PR for a new release; let me know how this looks to you and if you want to delay merging it until after you've wrapped up what you're currently doing.

mushtaqak:
@jmbowman We can not do a VAL release as of now due to https://github.com/edx/edx-platform/pull/17488 PR (which is code review now).

I will let you know when the corresponding PR is being merged and we are ready for a VAL release which I am planning after https://github.com/edx/edx-platform/pull/17488 is merged.

However, I think you can merge this PR but we can't release VAL as of now. Does this sound good to you?

jmbowman:
Ok, I'll remove the version bump from the PR and merge the rest.  An edx-val release with this is a prerequisite for doing an overhaul of the dependencies management in edx-platform, but there are a few other things I need to do in preparation for that anyway.  Thanks for the update!

mushtaqak:
@jmbowman We just released latest VAL version `0.1.12` :)

jmbowman:
Cool, and I see you already handled updating it in edx-platform. Thanks!

PR #137, opened 2018-05-02
moved delete transcript after create

irfanuddinahmad:


edxval/models.py
```diff
@@ -498,14 +498,14 @@ def create_or_update(cls, video, language_code, metadata, file_data=None):
                 setattr(video_transcript, prop, value)
 
         transcript_name = metadata.get('file_name')
+        try:
+            old_transcript_path = video_transcript.transcript.path
+        except ValueError:
+            old_transcript_path = None
```
muhammad-ammar: I think below is more cleaner

```python
if not created:
    old_transcript_path = video_transcript.transcript.path
```

edxval/models.py
```diff
@@ -518,6 +518,11 @@ def create_or_update(cls, video, language_code, metadata, file_data=None):
                     )
                     raise
 
+            # Delete the existing transcript file from storage(for example, AWS S3)
+            if not created and old_transcript_path is not None:
```
muhammad-ammar: we don't need to check for `old_transcript_path is not None` if we go with above comment.

edxval/tests/test_api.py
```diff
@@ -2480,6 +2480,37 @@ def test_create_or_update_video_exceptions(self, video_id, file_format, provider
...
         self.assertEqual(transcript_exception.exception.message, exception_message)
 
+    @mock.patch.object(VideoTranscript, 'save')
+    def test_create_or_update_transcript_atomicity(self, mock_save):
+        """
+        Verify `create_or_update_video_transcript` api function atomicity.
+        """
+        file_data = ContentFile(constants.TRANSCRIPT_DATA['overwatch'])
+        file_name = None
+        file_format = utils.TranscriptFormat.SRT
```
mushtaqak: let's not create extra variable here. We will just pass value to method directly

edxval/tests/test_api.py
```diff
@@ -2480,6 +2480,37 @@ def test_create_or_update_video_exceptions(self, video_id, file_format, provider
...
+                    file_format=file_format
+                ),
+                file_data=file_data
+            )
+
+        # Assert that there are no updates to the transcript data
+        video_transcript = VideoTranscript.objects.get(video__edx_video_id=edx_video_id,
+                                                       language_code=language_code)
+        with open(video_transcript.transcript.name) as saved_transcript:
+            self.assertNotEqual(saved_transcript.read(), constants.TRANSCRIPT_DATA['overwatch'])
```
mushtaqak: We can also verify the path is correct one (i.e it hasn't changed)

edxval/tests/test_api.py
```diff
@@ -2480,6 +2480,37 @@ def test_create_or_update_video_exceptions(self, video_id, file_format, provider
 
         self.assertEqual(transcript_exception.exception.message, exception_message)
 
+    @mock.patch.object(VideoTranscript, 'save')
+    def test_create_or_update_transcript_atomicity(self, mock_save):
```
mushtaqak: Do we need a test where we don't get the error while saving? (i.e transcript is updated and old file is deleted correctly)

edxval/tests/test_api.py
```diff
@@ -2663,7 +2694,6 @@ def test_no_create_transcript_file(self, video_id, language_code):
         # Verify no file is created.
         self.assertEqual(file_system.listdir(constants.EXPORT_IMPORT_STATIC_DIR), [])
 
-
```
mushtaqak: unintended line removal? This should not be removed.

PR #167, opened 2019-01-09
Missing HLS profile view accepts unlimited course IDs

Qubad786:
Missing HLS profile view accepts unlimited course IDs
 -- changed GET to POST and fixed corresponding tests

### DIFF
--- a/edxval/views.py
+++ b/edxval/views.py
@@ -268,10 +268,41 @@ class HLSMissingVideoView(APIView):
     """
     authentication_classes = (OAuth2Authentication, SessionAuthentication)
 
-    def get(self, request):
-        courses = request.query_params.getlist('courses')
-        batch_size = int(request.query_params.get('batch_size', 50))
-        offset = int(request.query_params.get('offset', 0))
+    def post(self, request):
+        """
+        Retrieve video IDs that are missing HLS profiles. This endpoint supports 2 types of input data:
+
+        1. If we want a batch of video ids which are missing HLS profile irrespective of their courses, the request
+           data should be in following format:
+                {
+                    'batch_size': 50,
+                    'offset': 0
+                }
+           And response will be in following format:
+                {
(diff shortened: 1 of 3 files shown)


edxval/views.py
```diff
@@ -268,10 +268,10 @@ class HLSMissingVideoView(APIView):
...
     authentication_classes = (OAuth2Authentication, SessionAuthentication)
 
-    def get(self, request):
-        courses = request.query_params.getlist('courses')
-        batch_size = int(request.query_params.get('batch_size', 50))
-        offset = int(request.query_params.get('offset', 0))
+    def post(self, request):
+        courses = request.data.get('courses')
+        batch_size = int(request.data.get('batch_size', 50))
+        offset = int(request.data.get('offset', 0))
```
muhammad-ammar: why we need to convert these to `int`, shouldn't they already be in integer format?

edxval/views.py
```diff
@@ -268,10 +268,10 @@ class HLSMissingVideoView(APIView):
     """
     authentication_classes = (OAuth2Authentication, SessionAuthentication)
 
-    def get(self, request):
-        courses = request.query_params.getlist('courses')
-        batch_size = int(request.query_params.get('batch_size', 50))
-        offset = int(request.query_params.get('offset', 0))
+    def post(self, request):
```
muhammad-ammar: Please add docstring to explain what we are doing in this POST?

setup.py
```diff
@@ -43,7 +43,7 @@ def load_requirements(*requirements_paths):
     return list(requirements)
 
 
-VERSION = '0.1.25'
+VERSION = '0.1.26'
```
muhammad-ammar: `GET` to `POST` and `POST` to `PUT` are breaking changes, and as per semantic versioning we should increase the MAJOR version which means `0.1.25` -> `1.1.25`

Qubad786:
@muhammad-ammar updated.

PR #179, opened 2019-08-07
python 2 to python 3 conversion

jinder1s:

PR #180, opened 2019-08-12
Test on python 3.5 and don't allow failures.

feanil:


feanil:
Nevermind I see that this is still in progress over here: https://github.com/edx/edx-val/pull/179/files

PR #181, opened 2019-08-19
Fix python3 incompatibilities

davestgermain:


edxval/api.py
```diff
@@ -1051,7 +1051,7 @@ def import_from_xml(xml, edx_video_id, resource_fs, static_dir, external_transcr
 
         return edx_video_id
     except ValidationError as err:
-        logger.exception(err.message)
+        logger.exception(xml)
```
feanil: Is the error message not useful? Can we provide some thing other than just the raw XML?

edxval/tests/test_serializers.py
```diff
@@ -98,8 +98,8 @@ def test_invalid_course_id(self):
         )
         self.assertFalse(serializer.is_valid())
         self.assertEqual(
-            serializer.errors,
-            {"courses": ["Ensure this value has at most 255 characters (it has 300)."]}
+            str(serializer.errors['courses']['course_id'][0]),
```
feanil: This seems like a different error structure than previously.  Did this change because of an upstream requirment changing or somethnig else?

edxval/api.py
davestgermain: It's redundant. `logger.exception()` will already include the exception message in the stacktrace. The argument is just extra info to include.

edxval/tests/test_serializers.py
davestgermain: This seems to be related to inadvertently upgrading django-rest-framework, which changes the error message representation. I'll revert this.

PR #185, opened 2019-10-01
PROD-776: Update authentication_classes to use JWTAuthentication

azarembok:
This is needed so that veda can authenticate with DOT auth.

### DIFF
--- a/edxval/views.py
+++ b/edxval/views.py
@@ -11,7 +11,8 @@ from rest_framework.authentication import SessionAuthentication
 from rest_framework.permissions import DjangoModelPermissions
 from rest_framework.response import Response
 from rest_framework.views import APIView
-from rest_framework_oauth.authentication import OAuth2Authentication
+
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 
 from edxval.api import create_or_update_video_transcript
 from edxval.models import (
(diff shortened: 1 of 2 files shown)


edxval/views.py
```diff
@@ -75,7 +77,7 @@ class VideoList(generics.ListCreateAPIView):
     """
     GETs or POST video objects
     """
-    authentication_classes = (OAuth2Authentication, SessionAuthentication)
+    authentication_classes = (JwtAuthentication, OAuth2Authentication, SessionAuthentication)
```
doctoryes: Why is the `OAuth2Authentication` being left in this list? Isn't `JwtAuthentication` sufficient?

azarembok: @doctoryes I mistakenly thought it was still required for the tests to pass, but it turns out that they only need SessionAuthentication to be there.  So I have now removed OAuth2Authentication.

timmc-edx:
This appears to have been the last reference to `rest_framework_oauth` in this repo—does that mean that the `djangorestframework-oauth` dependency can be removed now?

azarembok:
> This appears to have been the last reference to `rest_framework_oauth` in this repo—does that mean that the `djangorestframework-oauth` dependency can be removed now?

Yes, it looks like that can be removed.

PR #192, opened 2019-12-03
BOM-1043

awais786:
-Improved Quality
-Upgraded Requirements
-Fixed tests for Django22


Ayub-Khan:
@jmbowman I think this pr is ready to go. Should we bump version or not?

PR #194, opened 2019-12-05
BOM Project Switch to pytest

Ayub-Khan:
-Updated tox and setup.cfg
-Updated requirements
-remove django-nose
-minor fix in Makefile

### DIFF
--- a/edxval/tests/test_api.py
+++ b/edxval/tests/test_api.py
@@ -174,7 +174,7 @@ class CreateVideoTest(TestCase):
         }
         edx_video_id = api.create_external_video(display_name=expected_video['client_video_id'])
         video = VideoSerializer(Video.objects.get(edx_video_id=edx_video_id)).data
-        self.assertDictContainsSubset(expected_video, video)
+        assert expected_video == {k: v for k, v in video.items() if k in expected_video}
 
 
 @ddt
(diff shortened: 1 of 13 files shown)


edxval/tests/test_api.py
```diff
@@ -174,7 +174,8 @@ def test_create_external_video(self):
         }
         edx_video_id = api.create_external_video(display_name=expected_video['client_video_id'])
         video = VideoSerializer(Video.objects.get(edx_video_id=edx_video_id)).data
-        self.assertDictContainsSubset(expected_video, video)
+        for item in expected_video.items():
+            assert item in video.items()
```
jmbowman: The original assertion also checked that the values match in both dicts for keys in the subset dict.  I've seen a few different suggestions for replacing this deprecated method, I think the cleanest I've seen so far is at https://github.com/pytest-dev/pytest/issues/2376#issuecomment-496661974

edxval/tests/test_serializers.py
```diff
@@ -127,7 +127,8 @@ def test_encoded_video_set_output(self):
         # Check for 3 EncodedVideo entries
         self.assertEqual(len(result.get("encoded_videos")), 3)
         # Check for original Video data
-        self.assertDictContainsSubset(constants.VIDEO_DICT_FISH, result)
+        for item in constants.VIDEO_DICT_FISH.items():
+            assert item in result.items()
```
jmbowman: Same note on the deprecated assertion applies here.

edxval/tests/test_api.py
Ayub-Khan: updated. items() returns the tuple (key, value) and when comparing both values are checked.

PR #195, opened 2019-12-05
DE-1824: step 2 of 4 in renaming 'exists' field in ThirdPartyTranscriptCredentialsState to 'has_creds' (v1.1.33)

brianhw:
Add migration: copy 'exists' values to 'has_creds'.


setup.py
```diff
@@ -46,7 +46,7 @@ def load_requirements(*requirements_paths):
     return list(requirements)
 
 
-VERSION = '1.1.31'
+VERSION = '1.1.32'
```
brianhw: I'll update this to 1.1.33, since 1.1.32 has already been taken.

edxval/models.py
```diff
@@ -660,7 +660,7 @@ class Meta:
         choices=TranscriptProviderType.CHOICES,
     )
     # TODO remove exists in step 3 of renaming.
-    exists = models.BooleanField(default=False, help_text=u'Transcript credentials state')
+    exists = models.NullBooleanField(default=False, help_text=u'Transcript credentials state')
```
brianhw: I needed to change this type so that unit tests in step 3 would be able to still update the database column after the value-setting logic is removed from constructor methods.   It seems that the default value on the BooleanField was not enough to provide a value when saving new rows in the unit tests, at least on Sqllite.  So in order to be able to write NULL values to the column, the column must be nullable.

doctoryes: This change makes me worry about potential rollback scenarios where the deploy occurs, the value somehow is filled with NULL, and a rollback is needed. Is it possible to change/remove the offending test instead? Or is there no chance of this scenario happening?

brianhw: Fair question.   All current rows have a value of True for this column, so it would only be an issue if new rows were added.  As of this step, the column would still be updated will non-null values in production.   Once step 3 lands, however, there would be an issue for new rows added between step 3 and step 4 were one to roll back to step 1.  My understanding is that once step 4 lands, we're no longer able to rollback in general because we've deleted the old data from the database.

brianhw: Another consideration:   there are 21 entries in total, three from 2019, the most recent being in June.   According to Alan, there are not any expected to be added in the near future.

brianhw: But just to amplify on my earlier reply to @doctoryes 's question, I believe that it is riskier to *not* change the type in step 2 than to do so.  It is not *just* a matter about SQLLite and unit tests, I believe.   The actual table is:

{code}
> describe edxval_thirdpartytranscriptcredentialsstate;
+----------+-------------+------+-----+---------+----------------+
| Field    | Type        | Null | Key | Default | Extra          |
+----------+-------------+------+-----+---------+----------------+
| id       | int(11)     | NO   | PRI | NULL    | auto_increment |
| created  | datetime(6) | NO   |     | NULL    |                |
| modified | datetime(6) | NO   |     | NULL    |                |
| org      | varchar(32) | NO   | MUL | NULL    |                |
| provider | varchar(20) | NO   |     | NULL    |                |
| exists   | tinyint(1)  | NO   |     | NULL    |                |
+----------+-------------+------+-----+---------+----------------+
{code}
So `exists` is not nullable and has a default value of NULL.   The default=False that is specified at the model level does not exist at the database level.  So in step 3, once `exists` is removed from the model, there is nothing to provide a non-default value to a non-nullable database column with a NULL default, so production would also fail to work.

In terms of rollback, once step 3 is committed, any new rows added will not be able to be rolled back correctly in general.   The new value is in the new column and not the old column, and there is no mechanism to put the new values back into the old column.

pwnage101: > In terms of rollback, once step 3 is committed, any new rows added will not be able to be rolled back correctly in general. The new value is in the new column and not the old column, and there is no mechanism to put the new values back into the old column.

This is a problem in general with the process that I can't quickly think of a good solution for, and haven't had a reason to dig deeper since most of our column changes were not to frequently updated tables (yet).

pwnage101: If you aren't sure, I would ask you to check that this actually causes the column to be nullable, since for django fields in general null=false is typically the default.

brianhw: According to doc for BooleanField:  "Changed in Django 2.1:  In older versions, this field doesn’t permit null=True, so you have to use NullBooleanField instead. Using the latter is now discouraged as it’s likely to be deprecated in a future version of Django."   So null=True is the whole purpose of the NullBooleanField.  I have also verified that a field in the LMS that is defined as NullBooleanField is indeed nullable at the database level.  (See certificates_certificategenerationcoursesetting.include_hours_of_effort.)

PR #196, opened 2019-12-06
DE-1824: step 3 of 4 in renaming 'exists' field in ThirdPartyTranscriptCredentialsState to 'has_creds' (v1.1.34)

brianhw:
The 'exists' column name in ThirdPartyTranscriptCredentialsState is a reserved keyword in SQL, so this PR continues the process of renaming the field to 'has_creds' in its place. This is step 3, which removes the logic that relies on 'exists'.    Step 1 added the new field (#191). Step 2 added a migration to copy the contents of existing 'exists' to 'has_creds', and also renamed the old column to a nullable field (#195).  Step 4 will be to add a migration to remove the 'exists' column from the database. This also requires bumping the version of edx-val, and updating edx-platform, which uses edx-val.


brianhw:
Not sure why the migrations are appearing in this, or why it reports conflicts despite rebasing.  But it will be rebased on a version master after step2 merges before this gets merged anyway, so all those issues should resolve.

edxval/models.py
```diff
@@ -659,20 +659,17 @@ class Meta:
         max_length=20,
         choices=TranscriptProviderType.CHOICES,
     )
-    # TODO remove exists in step 3 of renaming.
-    exists = models.BooleanField(default=False, help_text=u'Transcript credentials state')
```
pwnage101: >Not sure why the migrations are appearing in this, or why it reports conflicts despite rebasing

Perhaps your rebase was not successful?  This line should have been NullBooleanField if it was rebased.

brianhw:
Closed in favor of #198, which is a PR off of master.

PR #198, opened 2019-12-20
DE-1824: step 3 of 4 in renaming 'exists' field in ThirdPartyTranscriptCredentialsState to 'has_creds' (v1.2.1)

brianhw:
Remove code for 'exists' w/o migration.

Update version to 1.1.34.

Replaces #196


brianhw:
Note that while this can be merged to master, it should not be released until a corresponding PR to edx-platform that contains a disabling of a migration test.  Releasing this before then may result in it being pulled into someone else's "make upgrade PR on edx-platform, and that PR would fail.  Safer to wait to merge it and release it and "make upgrade" on edx-platform and make the PR in one pass, preferably when edx-platform releases are unfrozen.

PR #206, opened 2020-03-13
Add new status to VideoStatusView

zainab-amir:
In order to use `VideoStatusView` from vem, we need to add new status to the **VALID_VIDEO_STATUSES**

Status Added: 
`trancode_active`: Set after encoding jobs have started


DawoudSheraz:
Merged this PR. The version bump will be created on edx-platform.

PR #210, opened 2020-03-30
add transcript credentials api

DawoudSheraz:
### Description
This PR is adding two API endpoints for TranscriptCredentials model(added in https://github.com/edx/edx-val/pull/208) with the following capabilities:
-  GET the credentials for a given org & provider. This endpoint will be used by VEM. The read permission is restricted.
- Create/Update credentials for an org. This is an internal endpoint, utilized by platform.

The purpose of the API is to move the transcription information part out of VEDA and into the VAL as part of VEM development.  Most bits of the code are taken from VEDA with necessary changes https://github.com/edx/edx-video-pipeline/blob/master/VEDA_OS01/views.py#L119.

Another important bit that the following settings should be added to platform so that it can be overridden and eventually used by VAL during the deployment:
```
CIELO24_SETTINGS = dict(
    CIELO24_API_VERSION=1,
    CIELO24_BASE_API_URL="https://sandbox.cielo24.com/api",
    CIELO24_LOGIN_URL="https://sandbox.cielo24.com/api/account/login"
)
 ```

### DIFF
--- a/edxval/settings.py
+++ b/edxval/settings.py
@@ -179,3 +179,11 @@ VIDEO_TRANSCRIPTS_SETTINGS = dict(
 
 # Set this value in the environment-specific files (e.g. local.py, production.py, test.py)
 FERNET_KEYS = ['insecure-ferent-key']
+
+# Transcript provider settings variables, which will be overridden at deployment if needed
+# NOTE: These settings must be added in edx-platform settings to allow deployment override and val integration
+CIELO24_SETTINGS = dict(
+    CIELO24_API_VERSION=1,
+    CIELO24_BASE_API_URL="https://sandbox.cielo24.com/api",
+    CIELO24_LOGIN_URL="https://sandbox.cielo24.com/api/account/login"
+)
--- a/edxval/transcript_utils.py
+++ b/edxval/transcript_utils.py
@@ -16,7 +16,9 @@ from pysrt.srtexc import Error
 from six import text_type
 from six.moves import range
 
+from edxval.enum import TranscriptionProviderErrorType
 from edxval.exceptions import TranscriptsGenerationException
+from edxval.models import TranscriptProviderType
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: edxval/views.py [hunk 1]
```diff
@@ -373,3 +378,221 @@ def put(self, request):
...
+        except TranscriptCredentials.DoesNotExist:
+            status_code = status.HTTP_400_BAD_REQUEST
+            data = {'message': "Credentials not found for provider {provider} & organization {org}".format(
+                provider=provider,
+                org=org
+            )}
+
+        return Response(status=status_code, data=data)
+
+    def get_cielo_token_response(self, username, api_secure_key):
```
azarembok: I'm wondering if it would be cleaner to separate out provider-specific stuff like this (and the settings that go with it) into separate source files.  Just seems a little out of place to have Cielo-specific stuff here in the middle of views.py.  What do you think?

DawoudSheraz: Sure, it will make the view code more compact. I will update the PR.

DawoudSheraz: The settings should be left in Django settings. The reason being that Cielo API URL is different. Right now, we are using a sandboxing URL. In VEDA, the URL is defined in edx-video-worker [instance-config.yaml](https://github.com/edx/edx-video-worker/blob/master/instance_config.yaml#L53) but it is sandboxing URL in [edx-video-pipeline](https://github.com/edx/edx-video-pipeline/blob/master/instance_config.yaml#L56). That's why I am thinking that we can override Cielo URL settings at deployment and use actual API URL. Although I will make the code to use defaults if some settings aren't found.

edxval/settings.py
```diff
@@ -179,3 +179,11 @@
 
 # Set this value in the environment-specific files (e.g. local.py, production.py, test.py)
 FERNET_KEYS = ['insecure-ferent-key']
+
+# Transcript provider settings variables, which will be overridden at deployment if needed
+# NOTE: These settings must be added in edx-platform settings to allow deployment override and val integration
+CIELO24_SETTINGS = dict(
+    CIELO24_API_VERSION=1,
```
zainab-amir: nit: since these settings will live both on val/platform and vem, we should add a comment suggesting to change the version, whenever we do, in both places (VEM and VAL/Platform) or keep it in some config repo to update it in one place

DawoudSheraz:
@zainab-amir/ @azarembok Can you take a second look? I have done major code refactoring, mainly move the `post` part of API to be an internal API function(to comply with how edx-platform uses edxval). The tests have also been moved into different suites accordingly.

edxval/transcript_utils.py
```diff
@@ -162,3 +164,61 @@ def get_api_token(username, api_key):
...
+    """
+    error_message = None
+    missing = [attr for attr in attributes if attr not in credentials]
+    if missing:
+        error_message = '{missing} must be specified for {provider}.'.format(
+            provider=provider,
+            missing=' and '.join(missing)
+        )
+
+    return TranscriptionProviderErrorType.MISSING_REQUIRED_ATTRIBUTES, error_message
```
zainab-amir: Why is it always returning `MISSING_REQUIRED_ATTRIBUTES` error type? Can we return None if we don't set any error message? then we will not have to explicitly set error_type to None here
https://github.com/edx/edx-val/pull/210/commits/085200f496f0c5ac607ffaad379e06844fe33be9#diff-d78075b51c3378064d929ce6515f2181R1248

DawoudSheraz: right.

edxval/settings.py
DawoudSheraz: These settings aren't utilized by edxval as edxval is to be plugged in edx-platform. I will add the comment when adding the settings in platform.

DawoudSheraz: I added the settings here to show what set of values are present in it.

zainab-amir: okay

PR #214, opened 2020-04-10
Squash migrations for edxval.

dianakhuang:
https://openedx.atlassian.net/browse/ARCHBOM-1093


edxval/settings.py
```diff
@@ -187,3 +187,17 @@
     CIELO24_BASE_API_URL="https://sandbox.cielo24.com/api",
     CIELO24_LOGIN_URL="https://sandbox.cielo24.com/api/account/login"
 )
+
+# Required by Django 2.2 to run management commands.
+TEMPLATES = [
```
robrap, line 192: In the future, I'd get changes like this into a separate PR ahead of the squash migrations, so the squash migrations will contain nothing but the squash.

PR #216, opened 2020-04-13
Remove python2 testing, update requirements.

azarembok:
This PR removes the python2 tests from .travis.yml and tox.ini, as they are no longer needed.  Also ran `make upgrade` to update requirements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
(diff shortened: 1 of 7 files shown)


azarembok:
> It looks like you could also drop the constraints in `requirements/constraints.txt` on futures, more-itertools, and pytest.

👍 will do

PR #222, opened 2020-04-29
Add Error Description Field To Video Model

zainab-amir:
- Add new optional field in Video Model 
- Make changes to video update view to accept error description as a request param


adzuci:
Also, could you describe what the user impact will be if this table stays locked for a period of time and how many rows are in it in prod?

zainab-amir:
> Also, could you describe what the user impact will be if this table stays locked for a period of time and how many rows are in it in prod?

Total rows in prod are **604753**. We get ~100 videos per day, I think it won't be an issue

PR #224, opened 2020-05-08
transcript credentials ADR

DawoudSheraz:
### Description

ADR for transcript credentials migration to VAL

### DIFF
--- a/docs/decisions/0001-transcript-credentials-in-val.rst
+++ b/docs/decisions/0001-transcript-credentials-in-val.rst
@@ -1,0 +1,24 @@
+Transcript Credentials in VAL
+------------------------------
+
+Status
+=======
+Accepted
+
+Context
+========
+Transcript credentials are required to submit a transcription job, to a provider, for an user submitted video. They are unique per provider for an organization. The transcription workflow is not possible without these credentials.
+
+
+Decision
+=========
+With `VEDA`_, the credentials are located within a VEDA model. However, the other video data lives in VAL. This is a data locality inconsistency even though the credentials are required by VEDA only. With the plans to migrate to `VEM`_ from VEDA, the credentials would not be accessible by VEM if they were in VEDA. Therefore, it was decided to move the existing video transcript credentials from VEDA to VAL. This will bring the credentials in place with related video data. Furthermore, as VAL is accessible by other services(being part of platform), the credentials can be obtained by the services easily.
+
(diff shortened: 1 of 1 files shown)


DawoudSheraz:
> I think we should talk about why we decided not to just keep the credentials in VEM. Thats probably the first question people will ask...since VEM replaced VEDA, and the credentials are needed by VEM, why aren't they in VEM?

Right, I will update the PR

docs/decisions/0001-transcript-credentials-in-val.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Context
+========
+Transcript credentials are required to submit a transcription job, to a provider, for an user submitted video. They are unique per provider for an organization. The transcription workflow is not possible without these credentials.
+
+
+Decision
+=========
+With `VEDA`_, the credentials are located within a VEDA model. However, the other video data lives in VAL. This is a data locality inconsistency even though the credentials are required by VEDA only. With the plans to migrate to `VEM`_ from VEDA, the credentials would not be accessible by VEM if they were in VEDA. Therefore, it was decided to move the existing video transcript credentials from VEDA to VAL. This will bring the credentials in place with related video data. Furthermore, as VAL is accessible by other services(being part of platform), the credentials can be obtained by the services easily.
+
```
zainab-amir, line 16: > However, the other video data lives in VAL

shouldn't it be course related data? or at least we should mention that course objects and details are in VAL and we didn't want to duplicate those and also wanted to keep VEM solely for the purpose of encoding
`other video data` might confuse readers as to why "video" data is not in VEM

DawoudSheraz: > > However, the other video data lives in VAL
> 
> shouldn't it be course related data? or at least we should mention that course objects and details are in VAL and we didn't want to duplicate those and also wanted to keep VEM solely for the purpose of encoding
> `other video data` might confuse readers as to why "video" data is not in VEM

VAL doesn't really have a course/org associated data like VEDA. Instead, it has CourseVideo that relates video in a course(via course_id). I will update the ADR to make the statement more specific.

DawoudSheraz:
Holding off work until we have a in-depth discussion on why this change was done in the first place.

DawoudSheraz:
Given the team's internal deliberation, I am closing this PR. This ADR is not valid anymore as the changes did will be reverted soon.

PR #229, opened 2020-05-18
Remove transcript credentials and preferences view

zainab-amir:
With our decision to move the transcript preferences and credential model to VEM, we no longer need following in edx-val:
* TranscriptPreferenceView
* TranscriptCredentialsView
* TranscriptCredentials model

Ticket: [PROD-1597](https://openedx.atlassian.net/browse/PROD-1597)

### NOTE:
Currently `TranscriptCredentials` model has only 21 db records in production


DawoudSheraz:
Remember to bump the version in setup.py before merging.

zainab-amir:
Devops Ticket for review: https://openedx.atlassian.net/browse/DOS-1140

Version will be updated in next PR, where we remove the `django-fernet-fields` from val requirements

edxval/migrations/0003_delete_transcriptcredentials.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('edxval', '0002_add_error_description_field'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='TranscriptCredentials',
+        ),
```
adzuci, line 15: Do you need to do this in two PRs, one for the code change and one for the migration?

Also, could we add a reverse migration to put this back just in case / to follow best practices?

zainab-amir: I will update the PR to separate code changes and migration changes

PR #255, opened 2020-08-05
course export with unpublished transcript fix

Ali-D-Akbar:
### Description
Exporting a course while having a video transcript that is yet to be published causes the export to fail. Exporting works fine in case of a published transcript file.

### Fix:
The lack of directories in the file system causes this issue to occur. There is a static subdirectory that needs to be made before creating the transcript file. So if the directory doesn't exist, it must be made in the File System.

### Reviewers
- [x] @DawoudSheraz 
- [ ] @azarembok 

### Post Review
 - [x] Squash & Rebase commit

### DIFF
--- a/edxval/tests/test_api.py
+++ b/edxval/tests/test_api.py
@@ -2885,6 +2885,40 @@ class TranscriptTest(TestCase):
         transcript = api.get_video_transcript_data(video_id=video_id, language_code=language_code)
         self.assertEqual(transcript['content'], expected_transcript_content)
 
+    def test_unpublished_create_transcript_file(self):
+        """
+        Tests that an unpublished transcript file is created correctly.
+
+        Scenario:
+        In a course with unpublished transcript, static_dir directory isn't made
+        so we don't explicitly make static_dir in this test
+        exporting the transcript will make static_dir inside file_system if not present
+        """
+        language_code = 'en'
+        video_id = constants.VIDEO_DICT_FISH['edx_video_id']
+        transcript_file_name = 'super-soaker-en.srt'
+        expected_transcript_path = combine(
+            combine(self.temp_dir, constants.EXPORT_IMPORT_COURSE_DIR),
+            combine(constants.EXPORT_IMPORT_STATIC_DIR, transcript_file_name)
+        )
+
+        delegate_fs = OSFS(self.temp_dir)
+        file_system = delegate_fs.makedir(constants.EXPORT_IMPORT_COURSE_DIR, recreate=True)
+
+        api.create_transcript_file(
+            video_id=video_id,
+            language_code=language_code,
+            file_format=utils.TranscriptFormat.SRT,
+            static_dir=constants.EXPORT_IMPORT_STATIC_DIR,
+            resource_fs=file_system
+        )
+
+        self.assertIn(transcript_file_name, file_system.listdir(constants.EXPORT_IMPORT_STATIC_DIR))
+
+        expected_transcript_content = File(open(expected_transcript_path, 'rb')).read()
+        transcript = api.get_video_transcript_data(video_id=video_id, language_code=language_code)
+        self.assertEqual(transcript['content'], expected_transcript_content)
+
     @data(
         ('invalid-video-id', 'invalid-language-code'),
         ('super-soaker', 'invalid-language-code')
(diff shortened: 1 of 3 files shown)


edxval/tests/test_api.py
```diff
@@ -2885,6 +2885,39 @@ def test_create_transcript_file(self):
         transcript = api.get_video_transcript_data(video_id=video_id, language_code=language_code)
         self.assertEqual(transcript['content'], expected_transcript_content)
 
+    def test_unpublished_create_transcript_file(self):
+        """
+        Tests that an unpublished transcript file is created correctly.
+        """
+        language_code = 'en'
+        video_id = constants.VIDEO_DICT_FISH['edx_video_id']
+        transcript_file_name = u'super-soaker-en.srt'
```
DawoudSheraz, line 2894: no need for `u`

edxval/tests/test_api.py
```diff
@@ -2885,6 +2885,39 @@ def test_create_transcript_file(self):
...
+        expected_transcript_path = combine(
+            combine(self.temp_dir, constants.EXPORT_IMPORT_COURSE_DIR),
+            combine(constants.EXPORT_IMPORT_STATIC_DIR, transcript_file_name)
+        )
+
+        delegate_fs = OSFS(self.temp_dir)
+        file_system = delegate_fs.makedir(constants.EXPORT_IMPORT_COURSE_DIR, recreate=True)
```
DawoudSheraz, line 2901: why explicit create here?

edxval/tests/test_api.py
```diff
@@ -2885,6 +2885,39 @@ def test_create_transcript_file(self):
...
+        # static_dir isn't made when the transcript is not published
```
DawoudSheraz, line 2902: Instead of inline comments, explain the scenario in unit test docstring

edxval/tests/test_api.py
Ali-D-Akbar: removing it.

edxval/tests/test_api.py
Ali-D-Akbar: I'll move it into the docstring.

edxval/tests/test_api.py
Ali-D-Akbar: it is a temporary course directory that is always made before export the transcript.
EXPORT_IMPORT_COURSE_DIR is an OSFS object having a directory something like `/tmp/tmpbo814ql3` and in this directory, `create_transcript_file` will make a `static_dir` in case of an unpublished transcript.

(`static_dir` is `course/static` in this case)

edxval/tests/test_api.py
```diff
@@ -2885,6 +2885,43 @@ def test_create_transcript_file(self):
...
+            file_format=utils.TranscriptFormat.SRT,
+            static_dir=constants.EXPORT_IMPORT_STATIC_DIR,
+            resource_fs=file_system
+        )
+
+        # Verify transcript file is created.
+        self.assertIn(transcript_file_name, file_system.listdir(constants.EXPORT_IMPORT_STATIC_DIR))
+
+        # Also verify the content of created transcript file.
+        expected_transcript_content = File(open(expected_transcript_path, 'rb')).read()
```
DawoudSheraz, line 2921: remove internal comments

edxval/tests/test_api.py
```diff
@@ -2885,6 +2885,43 @@ def test_create_transcript_file(self):
...
         self.assertEqual(transcript['content'], expected_transcript_content)
 
+    def test_unpublished_create_transcript_file(self):
+        """
+        Tests that an unpublished transcript file is created correctly.
+
+        Scenario:
+        In a course with unpublished transcript, static_dir directory isn't made
+        so we don't explicitly make static_dir in this test
+        create_transcript_file will make static_dir inside file_system
```
DawoudSheraz, line 2895: don't mention exact function names here. Just give context that directory will be created if not present etc.

DawoudSheraz:
Please update the version in setup.py before merging.

edxval/tests/test_api.py
Ali-D-Akbar: internal comments have been removed.

edxval/tests/test_api.py
Ali-D-Akbar: added an improved docstring.

PR #258, opened 2020-08-20
update transcript directory structure during export

DawoudSheraz:
### Description
VAL transcript export in static always expects directory structure to be `course/static`. This is valid for new mongo courses but for old mongo courses, the sub-directory is not `course` but based on the course run. This resulted in export failure for old mongo courses if the transcript was not published. This fix is an enhancement to https://github.com/edx/edx-val/pull/255 which fixes the issue for new courses but not old ones.

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -974,6 +974,19 @@ def create_transcripts_xml(video_id, video_el, resource_fs, static_dir):
     if video_transcripts.exists():
         transcripts_el = SubElement(video_el, 'transcripts')
 
+    # Create static directory based on the file system's subdirectory,
+    # falling back to default path in case of an error
+    try:
+        # File system should not start from /draft directory.
+        static_file_dir = combine(resource_fs._sub_dir.split('/')[1], static_dir)  # pylint: disable=protected-access
+    except KeyError:
+        logger.exception(
+            "VAL Transcript Export: Error creating static directory path for video {} in file system {}".format(
+                video_id, resource_fs
+            )
+        )
+        static_file_dir = combine('course', static_dir)
+
     transcript_files_map = {}
     for video_transcript in video_transcripts:
         language_code = video_transcript.language_code
@@ -985,7 +998,7 @@ def create_transcripts_xml(video_id, video_el, resource_fs, static_dir):
                 language_code=language_code,
                 file_format=file_format,
                 resource_fs=resource_fs.delegate_fs(),
-                static_dir=combine(u'course', static_dir)  # File system should not start from /draft directory.
+                static_dir=static_file_dir
             )
             transcript_files_map[language_code] = transcript_filename
         except TranscriptsGenerationException:
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ def load_requirements(*requirements_paths):
     return list(requirements)
 
 
-VERSION = '1.4.1'
+VERSION = '1.4.2'
 
 if sys.argv[-1] == 'tag':
     print("Tagging the version on github:")


Ali-D-Akbar (approved):

edxval/api.py
```diff
@@ -974,6 +974,15 @@ def create_transcripts_xml(video_id, video_el, resource_fs, static_dir):
     if video_transcripts.exists():
         transcripts_el = SubElement(video_el, 'transcripts')
 
+    # Create static directory based on the file system's subdirectory,
+    # falling back to default path in case of an error
+    try:
+        # File system should not start from /draft directory.
+        static_file_dir = combine(resource_fs._sub_dir.split('/')[1], static_dir)  # pylint: disable=protected-access
+    except Exception:  # pylint: disable=broad-except
```
kdmccormick, line 982: Is there a more specific exception type or set of exception types you could catch here?

edxval/api.py
```diff
@@ -974,6 +974,15 @@ def create_transcripts_xml(video_id, video_el, resource_fs, static_dir):
...
+        logger.exception("VAL Transcript Export: Error creating static directory path for video {}".format(video_id))
```
kdmccormick, line 983: Would it be useful to include anything like `static_dir` or `resource_fs` in this message?

DawoudSheraz:
@kdmccormick Thanks, I will address the comments

marcotuts:
Thanks for pushing this forward! Out of curiosity does VAL have release automation set up or does this need to be deployed manually? I'm curious when we could test whether this fix addresses old mongo export issues in production.

stvstnfrd:
> does VAL have release automation set up or does this need to be deployed manually? I'm curious when we could test whether this fix addresses old mongo export issues in production.

I _think_ we need a PR to bump the installed version in `edx-platform`.

edxval/api.py
```diff
@@ -974,6 +974,19 @@ def create_transcripts_xml(video_id, video_el, resource_fs, static_dir):
     if video_transcripts.exists():
         transcripts_el = SubElement(video_el, 'transcripts')
 
+    # Create static directory based on the file system's subdirectory,
+    # falling back to default path in case of an error
+    try:
+        # File system should not start from /draft directory.
+        static_file_dir = combine(resource_fs._sub_dir.split('/')[1], static_dir)  # pylint: disable=protected-access
```
swalladge, line 981: @DawoudSheraz it appears that it's possible for `resource_fs` to be a `WrapFS` instead of the expected `SubFS` at times, and `WrapFS` doesn't have the private `_sub_dir` attribute.  We're working on developing a fix now, but wondering if you had anything to add regarding the reason behind the particular combination of paths here (second path segment of the directory joined with the static_dir) and what kind of errors it should fallback on (what does KeyError catch here?)

DawoudSheraz: > @DawoudSheraz it appears that it's possible for `resource_fs` to be a `WrapFS` instead of the expected `SubFS` at times, and `WrapFS` doesn't have the private `_sub_dir` attribute. We're working on developing a fix now, but wondering if you had anything to add regarding the reason behind the particular combination of paths here (second path segment of the directory joined with the static_dir) and what kind of errors it should fallback on (what does KeyError catch here?)

Hi. What are the instances in the export where the resource_fs won't be SubFS? As for the join statements, the courses with old key format `org/run` don't have `course` as the root directory for export. It was the course run number. After the split, 2nd element in the list is the course run(which is the root directory for old-key format courses). For new courses, the same expression evaluates to `course`.

swalladge: Thanks @DawoudSheraz !  We're still trying to reproduce the issue, so aren't sure exactly yet. I _think_ it's related to one of the `override_export_fs` functions ([here](https://github.com/edx/edx-platform/blob/72aa58406b26f117a40fb259b1aee0f42e962e95/openedx/core/djangoapps/olx_rest_api/adapters.py#L100-L129) or [here](https://github.com/edx/edx-platform/blob/72aa58406b26f117a40fb259b1aee0f42e962e95/openedx/core/djangoapps/xblock/runtime/serializer.py#L96-L132). These both set a WrapFS filesystem, and are used in some functions that serialize xblocks ([here](https://github.com/edx/edx-platform/blob/72aa58406b26f117a40fb259b1aee0f42e962e95/openedx/core/djangoapps/xblock/runtime/serializer.py#L40) and [here](https://github.com/edx/edx-platform/blob/72aa58406b26f117a40fb259b1aee0f42e962e95/openedx/core/djangoapps/olx_rest_api/block_serializer.py#L83).  Perhaps it's only an issue when running the export olx api?

Do think think there will be a simple method that will support both WrapFS and SubFS (either using only public methods defined on both, or a fallback if `._sub_dir` is not available)?  I'm not familiar with how these virtual filesystems are used, so don't know the practical difference between the two for the purposes used here.

swalladge: @DawoudSheraz I investigated further and discovered that indeed a WrapFS is used for use with the olx_rest_api export method.  I opened https://github.com/edx/edx-val/pull/267 with a fix. :)

PR #268, opened 2020-11-08
[SE-3520] Adds waffle flag to enable overriding existing transcripts on import

nizarmah:
This pull-request is a follow up pull request, related to [this comment from a previous pull-request](https://github.com/edx/edx-val/pull/266#issuecomment-717822495).

Every time we import a new transcript, the existing and the new transcripts are compared by computing their content hash and comparing it.
This helps us identify whether the transcript uploaded is a duplicate of an already existing transcript, or if it is a different transcript that is being uploaded even though a transcript for the same video already exists.

Previously, if a video transcript for a specific video already exists, the new transcript wouldn't get uploaded. Now, if the transcript is different than the already existing one for a certain video, it does get uploaded and overrides it.

**JIRA tickets**: SE-3520, [OSPR-5117](https://openedx.atlassian.net/browse/OSPR-5117)

**Discussions**: [GitHub comment on previous pull-request](https://github.com/edx/edx-val/pull/266#issuecomment-717822495)

**Sandbox URL**:
- LMS: https://edxval-pr268.sandbox.opencraft.hosting/
- Studio: https://studio.edxval-pr268.sandbox.opencraft.hosting/

**Installation Instructions**:

- Install using:
```python
pip install --upgrade git+https://github.com/open-craft/edx-val.git@nizar/overriding_existing_non_duplicate_transcripts#egg=edxval
```

- Apply migrations to LMS
```bash
/edx/bin/edxapp-migrate-lms
```

**Testing instructions**:

- Testing using the Sandbox
    1. Login to the sandbox with `staff`/`edx`.
    1. Go to _Django Admin -> Waffle (django-waffle) -> Flag_ and activate the `edxval.override_existing_imported_transcripts` flag.
    1. Go to Studio and create a new course.
    1. Import one of the courses exported and attached below.
    1. Open the only unit in the course, edit the video, and download the transcript.
    1. Now import the other course, please.
    1. Open the only unit in the course, edit the video, and download the transcript.
    1. The transcript should be different than the other downloaded on step 4.
- Testing using Makefile
    1. Run `make requirements`
    1. Then run `make test`

> ##### Test Courses to Import
> - [First Course](https://github.com/edx/edx-val/files/5506797/course.rarp5q2h.tar.gz)
> - [Second Course](https://github.com/edx/edx-val/files/5506796/course.o5eu1225.tar.gz)

**Author notes and concerns**:

[...]

PR #279, opened 2021-01-26
[TSD] fixes toggle annotation name

nizarmah:
Addresses the [problem](https://github.com/edx/edx-val/pull/268/#issuecomment-767892293) with the waffle flag naming mentioned in https://github.com/edx/edx-val/pull/268.

Also, resolves the documentation problem for [`edxval.override_existing_imported_transcripts`](https://docs.google.com/spreadsheets/d/1xWbEL6oNu6D84WKBs3aViLRM40xk-vmnmmGAeLyR09A/edit?ts=6008a109&pli=1#gid=0&range=A178) in the doc-athon.

Also removes the python 3.5 function that was added because the repository was supported python 3.5 at that time.

### DIFF
--- a/edxval/config/waffle.py
+++ b/edxval/config/waffle.py
@@ -8,21 +8,7 @@ from edx_toggles.toggles.__future__ import WaffleFlag
 
 WAFFLE_NAMESPACE = 'edxval'
 
-
-def waffle_name(toggle_name):
-    """
-    Method to append waffle namespace to toggle's name
-
-    Reason behind not using f-strings is backwards compatibility
-    Since this is a method, it should be easy to change later on
-    """
-    return "{namespace}.{toggle_name}".format(
-        namespace=WAFFLE_NAMESPACE,
-        toggle_name=toggle_name,
-    )
-
-
-# .. toggle_name: OVERRIDE_EXISTING_IMPORTED_TRANSCRIPTS
+# .. toggle_name: edxval.override_existing_imported_transcripts
 # .. toggle_implementation: WaffleFlag
 # .. toggle_default: False
 # .. toggle_description: Enables overriding existing transcripts when importing courses with already
@@ -33,6 +19,5 @@ def waffle_name(toggle_name):
 # .. toggle_creation_date: 2021-01-01
 # .. toggle_tickets: https://openedx.atlassian.net/browse/OSPR-5117
 OVERRIDE_EXISTING_IMPORTED_TRANSCRIPTS = WaffleFlag(
-    waffle_name('override_existing_imported_transcripts'),
-    module_name=__name__,
+    f'{WAFFLE_NAMESPACE}.override_existing_imported_transcripts', __name__
 )


edx-webhook:
Thanks for the pull request, @nizarmah! I've created [OSPR-5564](https://openedx.atlassian.net/browse/OSPR-5564) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

edxval/config/waffle.py
```diff
@@ -8,21 +8,7 @@
...
-    Since this is a method, it should be easy to change later on
-    """
-    return "{namespace}.{toggle_name}".format(
-        namespace=WAFFLE_NAMESPACE,
-        toggle_name=toggle_name,
-    )
-
-
-# .. toggle_name: OVERRIDE_EXISTING_IMPORTED_TRANSCRIPTS
+# .. toggle_name: edxval.override_existing_imported_transcripts
```
robrap, line 11: @edx/core-committer-regisb: I’ve seen this issue elsewhere. Could we lint for “.” In the name for waffle flags and switches? Related, do we need to add the non namespaced version of these classes to the list of implementations for the annotation, or do we not want to add every class and subclass?

robrap (approved):
Looks good. Thank you!

natabene:
@nizarmah Thank you for your contribution. @robrap I see you are already on it!

edx-webhook:
@nizarmah 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/edx-val&entry.867055334=https://github.com/edx/edx-val/pull/279&entry.1484655318=https://github.com/nizarmah&entry.752974735=2021-01-26+23:21&entry.1917517419=2021-01-27+21:43&entry.2133058324=Yes) so we can improve your experience in the future.

robrap:
@nizarmah: Oops, I had removed "remove python 3.5 support" from the name of the PR and commit message because I thought that might be left-over from a different PR.  I now realize you were referring to the method that was removed.  Oh well. 

In any case, thank you again for the fix.

nizarmah:
@robrap It's alright! It shouldn't be an issue.
It's pretty clear now that python 3.5 support was removed since version 2.0.0 👍🏼 

I shouldn't have coupled those changes
But thanks for merging the changes anyway! 😃

edxval/config/waffle.py
regisb: > Could we lint for “.” In the name for waffle flags and switches?

Well, we could certainly do it, but then annotations would fail on [NonNamespaced*](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/waffle_utils/docs/decisions/0004-waffle-util-namespacing.rst#decision) objects, so I don't think we should do it.

> do we need to add the non namespaced version of these classes to the list of implementations for the annotation, or do we not want to add every class and subclass?

I think that the class used for the implementation is of no concern to the end user, so we should not be adding it.

robrap: @regisb: I agree, but still wish there was a solution to help avoid common problems. One idea would be to take the short flag name (i.e. everything after the period, if there even is one), and check to ensure the same appears in the name parameter of the call to the constructor? Just brainstorming for the future. This may also be risky and a bad idea.

PR #293, opened 2021-04-06
fix: allow conversion of Latin-1 encoded transcripts.

ormsbee:
```
﻿We try to always encode transcripts in UTF-8, but we have Latin-1
encoded files in older courses. So we have to fall back to that
when we get a UnicodeDecodeError (This is one of the causes of
export failures in TNL-8007.)
```


doctoryes (approved):
LGTM

ormsbee:
Thanks @doctoryes! I'll just patch the quality issues and merge. My devstack seems stuck at "quality installdeps". :-P

PR #294, opened 2021-04-09
Pact provider verification

DawoudSheraz:
### Description
Add provider verification mechanism in VAL. These changes don't include the build failure mechanism yet. The current VEM pact is bound to fail with oauth interaction. The major content for pact verification has been added.


edxval/pacts/utils.py
```diff
@@ -0,0 +1,129 @@
...
+Utility methods needed during pact verification.
+"""
+
+from edxval.models import CourseVideo, EncodedVideo, Profile, Video, VideoImage, VideoTranscript
+from edxval.tests import constants
+from edxval.views import VideoDetail, VideoImagesView, VideoStatusView, VideoTranscriptView
+
+VIDEO_ID = "3386eda1-8d6e-439d-tb89-e90a5c52h103"
+COURSE_ID = "course-v1:testX+test123+2030"
+DEFAULT_PROFILE_NAMES = ['desktop_mp4', 'mobile_low', 'mp3', 'hls']
```
azanbinzahid, line 11: Is there a single source for this list?

edxval/urls.py
```diff
@@ -33,3 +34,15 @@
         name='update-video-images'
     ),
 ]
+
+try:
+    # Add this url only if the test-only django settings is present
+    settings.TEST_ONLY_URLS  # pylint: disable=pointless-statement
```
azanbinzahid, line 40: Can be improved with this `getattr(settings, "TEST_ONLY_URLS", None)`

manage.py
```diff
@@ -5,7 +5,7 @@
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "edxval.settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "edxval.settings.base")
```
azanbinzahid, line 8: how it was handling these settings before?

azanbinzahid (approved):
Left a few comments to address, slightly uncertain if this is the right approach but as per discussion, we can go forward with the change as an initial attempt to provider-side testing.

manage.py
DawoudSheraz: there were no different settings file. Only settings.py. I created different settings file and base.py is same as settings.py

edxval/pacts/utils.py
DawoudSheraz: In migration https://github.com/edx/edx-val/blob/master/edxval/migrations/0001_squashed_0016_add_transcript_credentials_model.py#L18. and it has some extra profiles not used by VEM.

edxval/pacts/views.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    state_setup_mapping = {
+        'A valid video_id video exists': setup_successful_video_status_state,
+        'A valid video with no details exists': setup_successful_video_details_state,
+        'A valid video with no image information exists': setup_successful_video_image_state,
+        'A valid video with no video transcript information exists': setup_successful_video_transcripts_state,
+        'A valid video and video transcript information exists': setup_unsuccessful_video_transcripts_state,
+        'No video and video details exist': setup_unsuccessful_video_details_state,
+        'No valid video with video images details exist': setup_unsuccessful_video_image_state,
+        'A video against given video id does not exist': setup_unsuccessful_video_status_state,
```
Ali-D-Akbar, line 37: So, these are the only states that we need to test on the consumer side, right? So whenever there's a new state in provider_state, we might have to write a new test on consumer. Let me know if I was wrong anywhere.

Ali-D-Akbar (approved):
Just a question. Otherwise, LGTM.

edxval/pacts/views.py
DawoudSheraz: you got it reversed. Provider state identifies what condition the provider should be in before starting the interaction(and to make sure the interactions are independent). If a new interaction is introduced to the consumer, its appropriate state may or may not be added here.
https://docs.pact.io/getting_started/provider_states

Ali-D-Akbar: Ohh makes sense.

PR #297, opened 2021-04-16
feat: add default auth middleware for pact verification

DawoudSheraz:
### Description
Adds an auth middleware(test-only) that adds authenticated user and bypass csrf for the requests during pact verification.

### DIFF
--- a/edxval/management/commands/README.md
+++ b/edxval/management/commands/README.md
@@ -1,0 +1,9 @@
+### Description
+
+verify_pact.py is the management command for the provider pact verification. It will be used to verify the pacts for which edx-val is the provider.
+Since pact provider verification requires a provider url where the interactions will be made, the server should be running before 
+executing the management command. Therefore,
+
+1. Run ``python manage.py migrate --settings=edxval.settings.test`` to run the migrations for test database. This will be one-time step for local testing, unless DB structure is changed that will affect Pact verification.
+2. Run ``python manage.py runserver --settings=edxval.settings.test`` to start the server against which Pact provider verification will take place.
+3. Run ``python manage.py verify_pact --settings=edxval.settings.test`` in another terminal to start the pact verification process.
--- a/edxval/pacts/utils.py
+++ b/edxval/pacts/utils.py
@@ -4,7 +4,6 @@ Utility methods needed during pact verification.
 
 from edxval.models import CourseVideo, EncodedVideo, Profile, Video, VideoImage, VideoTranscript
 from edxval.tests import constants
-from edxval.views import VideoDetail, VideoImagesView, VideoStatusView, VideoTranscriptView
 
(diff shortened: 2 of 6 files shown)


edxval/management/commands/README.md
```diff
@@ -0,0 +1,8 @@
+### Description
+
+verify_pact.py is the management command for the provider pact verification. It will be used to verify the pacts for which edx-val is the provider.
+Since pact provider verification requires a provider url where the interactions will be made, the server should be running before 
+executing the management command. Therefore,
+
+1. Run ``python manage.py runserver --settings=edxval.settings.test``. One time migrate command might be needed.
```
azanbinzahid, line 7: Please add migrate command with flags as step 1. Quite possible that a user like me run migrate without flag.

Ali-D-Akbar (approved):
Why are we deleting some utility functions? Do we not need it anymore? Other than that, I think it is good to go.

edxval/pacts/utils.py
```diff
@@ -24,66 +23,18 @@ def clear_database():
...
-    Setup the provider state for unsuccessful video image update.
-    """
-    remove_view_auth_and_perms(VideoImagesView)
-
-
 def setup_successful_video_details_state():
     """
     Setup the provider state for successful video details update call.
     """
-    remove_view_auth_and_perms(VideoDetail)
```
azanbinzahid, line 62: Q: Instead of this, I see we have a new middleware now, how its bypassing the auth?

edxval/settings/test.py
```diff
@@ -17,3 +17,6 @@
 TEST_ONLY_URLS = True
 PROVIDER_BASE_URL = 'http://127.0.0.1:8000'
 PROVIDER_STATES_URL = '{}/edxval/pact/provider_states/'.format(PROVIDER_BASE_URL)
+
+MIDDLEWARE = list(MIDDLEWARE) + ['edxval.pacts.middleware.AuthenticationMiddleware']
+MIDDLEWARE = tuple(MIDDLEWARE)
```
azanbinzahid, line 22: Can be improved to this: 
`MIDDLEWARE = MIDDLEWARE + ('edxval.pacts.middleware.AuthenticationMiddleware',)`

azanbinzahid (changes requested):

edxval/pacts/utils.py
DawoudSheraz: middleware is setting request.user to be an authenticated user instead of anonymous user.

azanbinzahid (approved):

DawoudSheraz:
> Why are we deleting some utility functions? Do we not need it anymore? Other than that, I think it is good to go.

because the previous util functions were removing auth per view basis. With the addition of a default auth user via middleware, the auth removal inside the provider state utils is no longer needed.

PR #325, opened 2021-11-04
build: convert pact verification to unit test framework method

DawoudSheraz:
### Description

- Convert edx-val pact provider verification in unit testing framework method to make it easier to integrate into CI
- Usage of environment variable in setting some pact configuration in test.py file
- Updating pytest to use **test** settings instead of base


setup.cfg
```diff
@@ -47,6 +47,6 @@ skip=
     edxval/wsgi.py
 
 [tool:pytest]
-DJANGO_SETTINGS_MODULE = edxval.settings.base
+DJANGO_SETTINGS_MODULE = edxval.settings.test
```
azanbinzahid, line 50: How are we providing django settings settings while running pact with `pytest`?

Ali-D-Akbar:
Can you update the readme for pact verification process with the updated steps to run?

setup.cfg
DawoudSheraz: like we did with LMS, by exporting the settings module env variable before executing the verify call.

Ali-D-Akbar (approved):
Forgot to confirm this review before. Just a question, the rest looks good to me.

edxval/settings/pact.py
```diff
@@ -0,0 +1,17 @@
...
+does not cause issue with regular unit testing.
+"""
+import os
+
+from edxval.settings.test import *  # pylint: disable=wildcard-import
+
+PUBLISH_VERIFICATION_RESULTS = os.environ.get('PUBLISH_VERIFICATION_RESULTS', 'False').lower() in ('true', 'yes', '1')
+PROVIDER_STATES_SETUP_VIEW_URL = True
+PROVIDER_BASE_URL = 'http://127.0.0.1:8000'
+PROVIDER_STATES_URL = '{}/edxval/pact/provider_states/'.format(PROVIDER_BASE_URL)
```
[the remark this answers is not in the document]
DawoudSheraz: Yes, nice catch.

ansabgillani (approved):

azanbinzahid (approved):

PR #326, opened 2021-11-08
test: add provider verification CI job

Ali-D-Akbar:
[PROD-2552](https://openedx.atlassian.net/browse/PROD-2552)

Adds a separate pact provider verification job in CI. The pacts are retrieved from pactflow, verification is performed and after that, the results are published with dev/prod tags and provider version.
Also, we're allowing [Pending Pacts](https://docs.pact.io/pact_broker/advanced_topics/pending_pacts/) in edx-val as not doing that may block the edxval from merging the new changes if one or more consumers pacts are failing (even when the provider is compatible with the deployed version of the consumer)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,4 +72,4 @@ jobs:
         export PUBLISH_VERSION=`git rev-parse --short HEAD`
         export PUBLISH_TAGS=${{ github.head_ref }}
         export PUBLISH_VERIFICATION_RESULTS=true
-        pytest edxval/pacts/verify_pact.py --ds=edxval.settings.pact
+        pytest -s edxval/pacts/verify_pact.py --ds=edxval.settings.pact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,35 @@ jobs:
       with:
         flags: unittests
         fail_ci_if_error: true
+
+
+  provider-verification:
+    name: Pact Provider Verification
+    runs-on: ubuntu-20.04
+    needs: run_tests
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: setup python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Install pip
+      run: pip install -r requirements/pip.txt
+
+    - name: Install Dependencies
+      run: |
+        pip install -r requirements/ci.txt
+        pip install -r requirements/test.txt
+
+    - name: Verify Pacts
+      run: |
+        export PACT_BROKER_BASE_URL='https://edx.pactflow.io'
+        export PACT_BROKER_TOKEN=${{ secrets.PACT_FLOW_ACCESS_TOKEN }}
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -41,3 +41,11 @@ jobs:
...
         flags: unittests
         fail_ci_if_error: true
+
+    - name: Publish Provider Verification Results
+      if: matrix.python-version == '3.8' && matrix.toxenv == 'django32'
+      run: |
+        export PACT_BROKER_BASE_URL=https://edx.pactflow.io
+        export PUBLISH_VERSION=`git rev-parse --short HEAD`
+        export PUBLISH_VERIFICATION_RESULTS=true
+        pytest edxval/pacts/verify_pact.py --ds=edxval.settings.pact
```
DawoudSheraz, line 51: This job should be run separately and not part of unit tests. In VEM, publish pact job is run with unit tests because the pact is created/updated during the unit test suite. The same is not true for verification. This job will read pact from the broker/local file and then run the verification.

.github/workflows/ci.yml
```diff
@@ -41,3 +41,34 @@ jobs:
       with:
         flags: unittests
         fail_ci_if_error: true
+
+
+  provider-verification:
+    name: Publish Provider Verification Results
```
DawoudSheraz, line 47: nit: Pact Provider Verification. No need to add publish in start.

.github/workflows/ci.yml
```diff
@@ -41,3 +41,34 @@ jobs:
...
+
+    - name: Install pip
+      run: pip install -r requirements/pip.txt
+
+    - name: Install Dependencies
+      run: |
+        pip install -r requirements/ci.txt
+        pip install -r requirements/test.txt
+
+    - name: Publish Results
```
DawoudSheraz, line 68: nit: verify pacts

.github/workflows/ci.yml
```diff
@@ -41,3 +41,34 @@ jobs:
...
+      run: |
+        export PACT_BROKER_BASE_URL='https://edx.pactflow.io'
+        export PACT_BROKER_TOKEN=${{ secrets.PACT_FLOW_ACCESS_TOKEN }}
+        export PUBLISH_VERSION=`git rev-parse --short HEAD`
+        export PUBLISH_VERIFICATION_RESULTS=true
+        pytest edxval/pacts/verify_pact.py --ds=edxval.settings.pact
```
DawoudSheraz, line 74: Add `-s` flag so that logs/console statements are visible.

Ali-D-Akbar: Wouldn't that diable all the capturing? https://docs.pytest.org/en/6.2.x/capture.html#setting-capturing-methods-or-disabling-capturing

DawoudSheraz: Oh. On local, using -s yielded in a lot more log output, so I thought it was giving more output. I didn't consult the documents.

DawoudSheraz (approved):
- To add a link to Pending pacts document in PR description and why it is enabled

azanbinzahid (approved):

PR #329, opened 2021-11-10
fix: use master/production on merge

Ali-D-Akbar:
The [previous job merge job](https://github.com/edx/edx-val/runs/4162168998?check_suite_focus=true) failed due to `github.head_ref` not giving `master` on merge.

Using the same method as we did in [VEM](https://github.com/edx/video-encode-manager/pull/291) using github.ref on merge action gives you the branch name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,9 +67,17 @@ jobs:
 
     - name: Verify Pacts
       run: |
+        export IS_MERGED=${{ github.event.pull_request.merged }}
         export PACT_BROKER_BASE_URL='https://edx.pactflow.io'
         export PACT_BROKER_TOKEN=${{ secrets.PACT_FLOW_ACCESS_TOKEN }}
         export PUBLISH_VERSION=`git rev-parse --short HEAD`
-        export PUBLISH_TAGS=${{ github.head_ref }}
+        if [ $IS_MERGED == false ]
+        then
+          export PUBLISH_TAGS=$GITHUB_HEAD_REF
+          export GIT_ENV='development'
+        else
+          export PUBLISH_TAGS=${GITHUB_REF:11}
+          export GIT_ENV='production'
+        fi
         export PUBLISH_VERIFICATION_RESULTS=true
         pytest -s edxval/pacts/verify_pact.py --ds=edxval.settings.pact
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -67,9 +70,17 @@ jobs:
...
         export PACT_BROKER_BASE_URL='https://edx.pactflow.io'
         export PACT_BROKER_TOKEN=${{ secrets.PACT_FLOW_ACCESS_TOKEN }}
         export PUBLISH_VERSION=`git rev-parse --short HEAD`
-        export PUBLISH_TAGS=${{ github.head_ref }}
+        if [ $IS_MERGED == false ]
+        then
+          export PUBLISH_TAGS=${{ github.head_ref }}
+          export GIT_ENV='development'
+        else
+          export PUBLISH_TAGS=${GITHUB_REF:11}
```
azanbinzahid, line 82: Why can't we use `github.ref` directly here?

ansabgillani: Same.

ansabgillani:
One nit to be answered mentioned by Azan. Otherwise, it's good.

ansabgillani (approved):

.github/workflows/ci.yml
Ali-D-Akbar: oh, it looks like we can use $GITHUB_REF directly as well. https://docs.github.com/en/actions/learn-github-actions/environment-variables

Removing the above declaration.

.github/workflows/ci.yml
```diff
@@ -67,9 +67,17 @@ jobs:
...
     - name: Verify Pacts
       run: |
+        export IS_MERGED=${{ github.event.pull_request.merged }}
         export PACT_BROKER_BASE_URL='https://edx.pactflow.io'
         export PACT_BROKER_TOKEN=${{ secrets.PACT_FLOW_ACCESS_TOKEN }}
         export PUBLISH_VERSION=`git rev-parse --short HEAD`
-        export PUBLISH_TAGS=${{ github.head_ref }}
+        if [ $IS_MERGED == false ]
+        then
+          export PUBLISH_TAGS=${{ github.head_ref }}
```
azanbinzahid, line 76: I believe we also have `GITHUB_HEAD_REF`, can we use it for sake of consistency?

azanbinzahid (changes requested):

azanbinzahid (approved):

PR #337, opened 2021-12-01
fix: use branch name directly

Jawayria:
$default-branch is only available in workflow-templates so we'll use branch name directly in this workflow

https://stackoverflow.com/questions/64781462/github-actions-default-branch-variable

### DIFF
--- a/.github/workflows/upgrade-python-requirements.yml
+++ b/.github/workflows/upgrade-python-requirements.yml
@@ -8,7 +8,7 @@ on:
        branch:
          description: 'Target branch to create requirements PR against'
          required: true
-         default: $default-branch
+         default: 'master'
 jobs:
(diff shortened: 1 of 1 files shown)


DawoudSheraz (commented):
what's the purpose of $default-branch if it is not usable in workflow? Won't this cause a lot of re-edits when the default branches are renamed to main?

aht007 (approved):

mraarif (approved):

Jawayria:
> what's the purpose of $default-branch if it is not usable in workflow? Won't this cause a lot of re-edits when the default branches are renamed to main?

Yes we'll have to make a lot of re-edits in that case. Actually this $default-branch placeholder is used in workflow templates so that when we add a workflow in repo using that workflow, it automatically gets replaced with default branch name.

PR #348, opened 2022-02-08
Cleanup django32

aht007:


.github/workflows/ci.yml
```diff
@@ -62,6 +62,7 @@ jobs:
 
     - name: Install Dependencies
       run: |
+        pip install "Django<4.0"
```
aht007, line 65: CI and test requirements file don't have Django as dependency in them since we remove Django from test file to be controlled by `tox` but for `pact verification` check we can't use tox as we can't move these commands to `tox` since they are using `github env` variables. This pin has been added due to this reason to constraint Django to less than 4.0 which otherwise would install Django4

PR #454, opened 2023-09-01
feat: add endpoint for course transcript details

BrandonHBodine:
Context: I'm working on a project that will give edX an automated path for translating transcripts using AI translations services.

This PR is to add an endpoint; that when passed a course_id will retrieve all it's related videos transcripts so that our service will be able to read the transcript files, translate them, and then update the translated transcripts back to VAL.


edxval/api.py
```diff
@@ -732,6 +732,49 @@ def get_videos_for_course(course_id, sort_field=None, sort_dir=SortDirection.asc
...
+
+    course_transcripts_data = {}
+
+    course_videos = CourseVideo.objects.filter(course_id=course_id)
+    for course_video in course_videos:
+
+        edx_video_id = course_video.video.edx_video_id
+        transcript_data = {}
+
+        video_transcripts = VideoTranscript.objects.filter(video=course_video.video)
```
DawoudSheraz, line 764: You can use Django's select_related (for 1-1, FK) and prefetch_related (reverse 1-M, M-M) to optimize DB fetch operations.

- https://docs.djangoproject.com/en/4.2/ref/models/querysets/#select-related
- https://docs.djangoproject.com/en/4.2/ref/models/querysets/#prefetch-related

Rodra (approved):

edxval/tests/test_views.py
```diff
@@ -1101,3 +1102,33 @@ def test_update_hls_encodes_for_video(self):
...
+            # Simulate a return value when the function is called.
+            mock_transcript_details.return_value = {}
+            course_id = 'course-v1:edx+1+2023_05'
+            url = reverse(self.base_url, args=[course_id])
+            response = self.client.get(url)
+
+            # Verify the function was called once with course_id
+            mock_transcript_details.assert_called_once_with(course_id)
+
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
```
DawoudSheraz, line 1134: nit: You should not mock the internal util here. Instead, setup the required data using factories and verify that view is getting the correct data as expected.

edxval/tests/test_views.py
```diff
@@ -1101,3 +1102,33 @@ def test_update_hls_encodes_for_video(self):
...
+
+class CourseTranscriptsDetailViewTest(APIAuthTestCase):
+    """
+    CourseTranscriptsDetailView Tests.
+    """
+    def setUp(self):
+        """
+        Tests setup.
+        """
+        self.base_url = 'course-transcripts'
```
DawoudSheraz, line 1115: nit: this can be moved to class level variable as this is constant.

DawoudSheraz (commented):
Please ensure that the version is bumped in https://github.com/openedx/edx-val/blob/master/edxval/__init__.py#L5. Once merged, the release will be created from https://github.com/openedx/edx-val/releases which pushes the latest version to PyPi. Once the latest version is available on Pypi, it will be updated in edx-platform.

edxval/tests/test_views.py
DawoudSheraz: @BrandonHBodine Hi, this was not addressed. I would suggest creating a followup PR to address this.

DawoudSheraz: Well, not a requirement for release but important from unit testing point of view. 
I agree that get_transcript_details_for_course has been tested in edxval/tests/test_api.py but the view is utilizing that util, so we have to ensure that view is working as expected by adding some test data. It is not redundant. Both are testing different parts of the code.

BrandonHBodine:
> Please ensure that the version is bumped in https://github.com/openedx/edx-val/blob/master/edxval/__init__.py#L5. Once merged, the release will be created from https://github.com/openedx/edx-val/releases which pushes the latest version to PyPi. Once the latest version is available on Pypi, it will be updated in edx-platform.

@DawoudSheraz Is this a manual process or is it automated somewhere? Do I create a new tag and draft a new release? I walked through that process but noticed that the other releases have assets, zip file and tar, included.

DawoudSheraz:
> > Please ensure that the version is bumped in https://github.com/openedx/edx-val/blob/master/edxval/__init__.py#L5. Once merged, the release will be created from https://github.com/openedx/edx-val/releases which pushes the latest version to PyPi. Once the latest version is available on Pypi, it will be updated in edx-platform.
> 
> @DawoudSheraz Is this a manual process or is it automated somewhere? Do I create a new tag and draft a new release? I walked through that process but noticed that the other releases have assets, zip file and tar, included.

You would need to draft a new release against master with an appropriate new version. Rest of that is automated.

PR #527, opened 2024-09-13
feat: remove transcript content from course-transcripts endpoint

BrandonHBodine:
The API endpoint we created included transcript content with it which we no longer need. This removes that data from the call.


hajorg (approved):
LGTM

PR #537, opened 2024-11-01
build: remove Python 3.8 references

DawoudSheraz:
Since Python 3.8 has reached the end of life, keeping 3.8 related references in edx-val is not needed.


AfaqShuaib09 (approved):

DawoudSheraz:
> It would be better to release a new version as well since the support of Python 3.8 is being dropped.

I will do it in a followup PR.

DawoudSheraz:
Created https://github.com/openedx/edx-val/pull/538 for version bump

PR #541, opened 2024-11-18
feat(AU-2283): Create transcript delete endpoint

Rodra:
https://2u-internal.atlassian.net/browse/AU-2283


edxval/views.py
```diff
@@ -410,3 +416,64 @@ def put(self, request):
...
+        data = request.data
+        missing_transcripts = []
+        for video_id, language_codes in data.items():
+            if not isinstance(language_codes, list):
+                return Response(
+                    status=status.HTTP_400_BAD_REQUEST,
+                    data={'message': f'Value for video "{video_id}" needs to be a list of language codes.'}
+                )
+            for language_code in language_codes:
+                if language_code not in get_available_transcript_languages(video_id=video_id):
```
jansenk, line 460: nit: it would be a slight improvement to only call `get_available_transcript_languages(video_id=video_id)` once rather than in the loop

jansenk (approved):

edxval/views.py
Rodra: Good catch

BrandonHBodine:
What kind of permission does a user need to have to hit this endpoint?

BrandonHBodine (changes requested):
Requesting changes so we don't open this endpoint up to non staff or non course authors.

BrandonHBodine (approved):

PR #543, opened 2024-11-25
docs: Add PII annotations

bmtcril:
Per OEP-30 these annotations should live with the models, just moving them here from the edx-platform safelist.


bmtcril:
@BrandonHBodine @jansenk I could use a quick review on this PR, but first do either of you have any idea why this is consistently failing the Pact tests?

bmtcril:
Closing this PR to open a new one that's not from my fork. Apparently forks don't have permissions to run the Pact tests.

PR #544, opened 2024-11-27
feat: add ai translations as a translation provider

jansenk:
Added `edx_ai_translations` as a transcript provider option

I didn't want it to also be an option for TranscriptPreference.provider or ThirdPartyTranscriptCredentialsState.provier at this point .... potentially in the future?


edxval/models.py
```diff
@@ -378,13 +378,20 @@ class for providing tuple choices.
...
     CIELO24 = 'Cielo24'
+    EDX_AI_TRANSLATIONS = 'edx_ai_translations'
 
     CHOICES = (
         (CUSTOM, CUSTOM),
         (THREE_PLAY_MEDIA, THREE_PLAY_MEDIA),
         (CIELO24, CIELO24),
     )
 
+    # Choices specififcally for only the VideoTranscript model
```
BrandonHBodine, line 389: Isn't the `CHOICES` enumeration already isolated to the VideoTranscript model?

BrandonHBodine (changes requested):
I don't think this will work without updating the line at https://github.com/openedx/edx-val/blob/11521dbeb85ffaca0022cce816c796558adc7a33/edxval/api.py#L385

edxval/models.py
jansenk: TranscriptPreference and ThirdPartyTranscriptCredentialsState also use this model

BrandonHBodine (approved):

PR #554, opened 2025-01-15
feat: create get_course_videos_qset API call

ormsbee:
This new call returns a CourseVideo queryset that will allow callers to
efficiently get what they want, instead of having to bring back a large
serialized object that may have many fields that they don't care about.

### DIFF
--- a/edxval/api.py
+++ b/edxval/api.py
@@ -733,6 +733,18 @@ def get_videos_for_course(course_id, sort_field=None, sort_dir=SortDirection.asc
     )
 
 
+def get_course_videos_qset(course_id):
+    """
+    Get a QuerySet of CourseVideos for a given course.
+
+    This assumes that the caller can further filter as necessary.
+    """
+    return CourseVideo.objects.select_related('video').filter(
+        course_id=str(course_id),
+        is_hidden=False,
+    )
+
+
 def get_transcript_details_for_course(course_id):
     """
     Get all the transcripts for a course and return details.
--- a/edxval/tests/test_api.py
+++ b/edxval/tests/test_api.py
@@ -761,6 +761,13 @@ class GetVideosForCourseTest(TestCase, SortedVideoTestMixin):
 
         self.assertEqual(len(course_transcript), 0)
 
+    def test_get_course_videos_qset(self):
+        """Test getting a minimal queryset."""
+        with self.assertNumQueries(1):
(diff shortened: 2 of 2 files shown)


ormsbee:
@jansenk, @BrandonHBodine: Hi folks. I'm creating this PR to enable https://github.com/openedx/edx-platform/pull/36112, which will make the course quality checks that load in the Studio outline page much faster for those courses that use edx-val extensively. I don't understand what's going on with the Pact stuff though.

Ali-D-Akbar (approved):

DawoudSheraz:
> @jansenk, @BrandonHBodine: Hi folks. I'm creating this PR to enable [openedx/edx-platform#36112](https://github.com/openedx/edx-platform/pull/36112), which will make the course quality checks that load in the Studio outline page much faster for those courses that use edx-val extensively. I don't understand what's going on with the Pact stuff though.

It seems the auth call to edx.pactflow.io is failing with 401. I don't have access to see the secret values on Github. Can you help check if PACT_FLOW_ACCESS_TOKEN is configured/present? Looking at edx.pactflow.io, the token was last used on January 13, so it is possible the config got removed from Open edX org secrets.

ormsbee:
@DawoudSheraz: I'll look into that, thank you!

DawoudSheraz:
> @DawoudSheraz: `PACT_FLOW_ACCESS_TOKEN` exists and this repo has access to it. Is it possible that the token expired and we need to put a new value for it?

Hmmm, strange. The token does not have an explicit expiry date.

DawoudSheraz:
Weird, I triggered the requirements upgrade flow job and pact CI is passing there https://github.com/openedx/edx-val/pull/555

DawoudSheraz:
I understand what is going on here. The PRs are being made using the fork, the direct push/write access is not available anymore. In that case, Github actions try to read the secret from fork's Github account rather than Open edX Github org secrets (https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow). Because the secret is not available in fork (obviously), pact is not able to authenticate with pactflow and raises 401.

ormsbee:
Huh. I guess I'll close this and re-open with a branch on this repo. That will be problematic for a lot of contributors though. FYI @feanil

ormsbee:
Closing because of Pact auth issues. Will re-open as a branch on this repo instead.

PR #557, opened 2025-01-17
feat: create get_course_videos_qset API call

ormsbee:
This new call returns a CourseVideo queryset that will allow callers to efficiently get what they want, instead of having to bring back a large serialized object that may have many fields that they don't care about.


ormsbee:
@DawoudSheraz: I'm just remaking https://github.com/openedx/edx-val/pull/554 from this repo instead of my fork, to get around the Pact issue.

DawoudSheraz (approved):

PR #562, opened 2025-02-06
feat: add patch endpoint to update transcript providers

BrandonHBodine:
This PR adds a DRF Patch endpoint for updating the "provider" field on VideoTranscript records.
Context:


edxval/views.py
```diff
@@ -217,6 +219,43 @@ def delete(self, request):
...
+                data={'message': 'Invalid transcript provider.'}
+            )
+
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            return Response(
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                data={'message': str(e)}
+            )
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
hajorg, line 257: Shouldn't this be 404 if the video transcript isn't found since `update_transcript_provider` will return None?

BrandonHBodine: Ah, yes that would be correct! Thanks for catching that.

edxval/api.py
```diff
@@ -395,6 +395,29 @@ def create_or_update_video_transcript(video_id, language_code, metadata, file_da
...
+        video_id: id identifying the video to which the transcript is associated.
+        language_code: language code of a video transcript.
+        provider: transcript provider
+    """
+    video_transcript = VideoTranscript.get_or_none(video_id, language_code)
+
+    if provider and provider not in list(dict(TranscriptProviderType.TRANSCRIPT_MODEL_CHOICES).keys()):
+        raise InvalidTranscriptProvider(f'{provider} transcript provider is not supported')
+
+    if video_transcript:
```
hajorg, line 412: While I know that empty/missing provider has been handled [here](https://github.com/openedx/edx-val/pull/562/files#diff-5cb93930cfbc5ddd56013cd73f08792eaee9fde2b7c8fc4a571bc609b12b665bR230). Should we make sure we are not passing an empty provider in this condition in case another function needs to use it in the future? I don't think [this](https://github.com/openedx/edx-val/pull/562/files#diff-5cb93930cfbc5ddd56013cd73f08792eaee9fde2b7c8fc4a571bc609b12b665bR230) will account for empty provider.

hajorg (approved):
Added a minor comment but LGTM otherwise

BrandonHBodine:
I was wondering if anyone from @openedx/committers-maintainers-at-large could review and merge this PR? My team is taking over maintainership of this repo soon but this PR is a blocker for some working being done. CC @AfaqShuaib09

UsamaSadiq:
Hi @BrandonHBodine 
I can help you in getting this PR merged but you'll need to resolve the conflict first.

BrandonHBodine:
The conflict has been resolved. Thank you for the help @UsamaSadiq

UsamaSadiq (approved):

UsamaSadiq:
Merging the PR without Pact Verification check passing since it is already failing in other PRs too.

BrandonHBodine:
@UsamaSadiq would you be able to create a release for this too? I also no longer have access to do so.

UsamaSadiq:
@BrandonHBodine I've released version [2.10.0](https://pypi.org/project/edxval/2.10.0/) on GitHub and PyPI with your change.

PR #577, opened 2025-04-14
Add support for Python 3.12 & Django 5.2

irtazaakram:
- resolves https://github.com/openedx/edx-val/issues/526
- resolves https://github.com/openedx/edx-val/issues/575

Major Change: https://github.com/openedx/public-engineering/issues/340


.github/workflows/ci.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
     - name: Install Dependencies
       run: |
-        pip install "Django>=4.2,<4.3"
+        pip install "Django>=4.2,<5.0"
```
awais786, line 66: No need to make change here.

tox.ini
```diff
@@ -1,9 +1,10 @@
 [tox]
-envlist = py{38,311}-django{42}, quality
+envlist = py{311,312}-django{42,52}, quality
 
 [testenv]
 deps =
-    django42: Django>=4.2,<4.3
+    django42: Django>=4.2,<5.0
```
awais786, line 6: Django>=4.2,<4.3
Django>=5.2,<5.3

awais786:
This change will affect the **edx-platform**, so we need to test this PR there as well before merging.

edxval/utils.py
```diff
@@ -155,13 +155,13 @@ def get_video_image_storage():
     Return the configured django storage backend.
     """
     if hasattr(settings, 'VIDEO_IMAGE_SETTINGS'):
-        return get_storage_class(
-            settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS'),
-        )(**settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_KWARGS', {}))
+        storage_name = settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS')
+        if storage_name and storage_name in storages:
+            return storages[storage_name]
```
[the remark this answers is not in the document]
mumarkhan999: Can you please check it while debugging.... and share the SS?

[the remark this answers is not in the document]
awais786: I have a question regarding **edx-platform**: if we merge this PR, will it work with the current platform settings, or will it require immediate updates?
Is it safe to merge now and handle any necessary setting changes when we update edx-platform later?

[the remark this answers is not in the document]
mumarkhan999: This PR requires `settings.py` to have `storages` dict in edx-platform as it will be installing inside edx-platform.
Which isn't available at the moment.

irtazaakram:
Created a PR here https://github.com/openedx/edx-platform/pull/36658 to test it in edx-platform.

edxval/utils.py
```diff
@@ -154,14 +154,8 @@ def get_video_image_storage():
...
     """
-    if hasattr(settings, 'VIDEO_IMAGE_SETTINGS'):
-        return get_storage_class(
-            settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS'),
-        )(**settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_KWARGS', {}))
-
-    # during edx-platform loading this method gets called but settings are not ready yet
-    # so in that case we will return default(FileSystemStorage) storage class instance
-    return get_storage_class()()
+    storage_name = settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS', 'default')
```
awais786, line 157: 1. The storages dictionary doesn't include a STORAGE_CLASS, so it falls back to the default.
2. Even if you add a STORAGE_CLASS, it tries to return the Boto3 storage class directly, which causes a crash—because storages expects an alias, not a class path.

```
(Pdb) storage_name = settings.VIDEO_IMAGE_SETTINGS.get('STORAGE_CLASS', 'default')
(Pdb) storage_name
'storages.backends.s3boto3.S3Boto3Storage'
(Pdb) storages[storage_name]
*** django.core.files.storage.handler.InvalidStorageError: Could not find config for 'storages.backends.s3boto3.S3Boto3Storage' in settings.STORAGES.
```

awais786: We need to find a solution which works with existing configurations.

edxval/settings/base.py
```diff
@@ -180,6 +176,15 @@
 )
 
 
+STORAGES = {
```
feanil, line 179: This is the default value in Django so I don't think you need to add it to our base settings file here.

edxval/utils.py
```diff
@@ -150,18 +150,37 @@ def video_image_path(video_image_instance, filename):  # pylint:disable=unused-a
...
 
 
+def get_configured_storage(settings_key):
+    """
+    Generic function to return a configured Django storage backend
+    based on the settings dictionary at `settings_key`.
+    """
+    config = getattr(settings, settings_key, {})
+    # Retrieve the storage class path and kwargs from the settings
+    storage_class_path = config.get('STORAGE_CLASS', {})
```
feanil, line 160: The class should not return an empty dict, it should return none like we used to previously in `get_video_image_storage`

edxval/utils.py
```diff
@@ -150,18 +150,37 @@ def video_image_path(video_image_instance, filename):  # pylint:disable=unused-a
     return '{}{}'.format(settings.VIDEO_IMAGE_SETTINGS.get('DIRECTORY_PREFIX', ''), filename)
 
 
+def get_configured_storage(settings_key):
+    """
+    Generic function to return a configured Django storage backend
+    based on the settings dictionary at `settings_key`.
```
feanil, line 156:
```suggestion
    Generic function to return a configured Django storage backend
    based on the settings dictionary at `settings_key`.  This function falls
    back to the `default` storage class if there is no `STORAGE_CLASS` entry
    under the `setting_key` object.
```

feanil (approved):

feanil:
Looks like you might need to rebuild the pact tests?

awais786:
@irtazaakram, since you're creating the PR from a fork, the Pact test is failing. Good to go.

PR #583, opened 2025-05-05
test: Adding tests for storages.

awais786:
This repository currently lacks some test cases for Django storages. I'm adding a few now to ensure better coverage, especially because an upcoming PR will remove some deprecated methods from Django storages. Adding these tests in advance will help us verify the behavior and impact of the removals.

Upcoming PR https://github.com/openedx/edx-val/pull/580


awais786:
Test is failing due to fork issue. As secret is missing for forks.

PR #592, opened 2025-10-14
feat: delete with provider validation

nsprenkle:
Adds an optional `provider` param for the `video-transcript` `DELETE` endpoint for additional validation.

```http
DELETE /val/v0/videos/video-transcripts?video_id={video_id}&language_code={language_code}&provider={provider} /HTTP/1.1
```

For backwards compatibility, this is _optional_: where omitted, the delete endpoint operates as previously defined and, whether found or not, returns a `204`.

```http
HTTP/1.1 204 No Content
```

Where provided (pun intended), the `provider` param checks that the given `provider` matches that of the identified transcript, failing with a `404`.

```http
HTTP/1.1 404 Not Found
Content-Type: application/json

{
    "message": "Transcript provider does not match, cannot delete the transcript."
}
```


michaelroytman (approved):
Looks good to me!

jansenk (approved):

PR #604, opened 2026-05-01
fix: missing transcript should not fail export

### LINKED ISSUE #60: Error with missing video transcripts
The demo course at the current master branch includes references to video transcript files which do not exist. This causes errors, such as failing to export the course or failing to reindex the course.

Example error:

> File does not exist: video-transcripts/93eb0b27dd8c48319fdb0258193958dd.srt

The references need to be removed or updated, or the missing files added.

## Reproducer

- download the demo-course.tar.gz from the dist directory on the master branch
- import the demo course into an Ulmo openedx instance (the issue is visible on older releases too, but this particular export error appears to be newly surfaced on Ulmo)
- export the course, and verify you see an error, and the course fails to export

## Workaround

A workaround for now is to delete the problematic video transcripts from LMS Django Admin -> Edxval -> Video transcripts.

Image

ormsbee:
Before this commit, a missing transcript file would raise a
FileNotFoundError from get_video_transcript_data() when the file was
missing. This can happen if there are references to edx-val video_ids in
the course content, but the corresponding transcript files were never
properly exported. This exception was never handled anywhere in edx-val,
so it would fail the course export entirely.

Now we catch the FileField's underlying FileNotFoundError and re-raise a
TranscriptNotFoundError, and then catch that a couple layers up in
create_transcripts_xml(). This means that a missing transcript will not
fail other transcripts for the same VideoBlock (though in practice, if
one is broken, the others are likely broken as well).

Fixes https://github.com/openedx/openedx-demo-course/issues/60


ormsbee:
I forgot that edx-val's Pact situation means that PRs can only be opened from this repo, not from a fork. :scream_cat: 

Closing this to re-open the same thing elsewhere.

PR #615, opened 2026-06-17
feat!: remove pact provider verification; drop Django 4.2; bump codecov

feanil:
## Summary

This PR consolidates several related cleanups originally motivated by trying to get the pact provider verification job working for fork PRs. Investigation revealed the local pact JSON (the only mode fork PRs could use, since they cannot access the `PACT_FLOW_ACCESS_TOKEN` secret) has not been updated since 2021-11-04, and the consumer-side source (`edx/video-encode-manager`) is now a private third-party repo with no workable sync path. Rather than maintain a slowly-rotting snapshot, the pact tooling is removed entirely.

While here, bump codecov-action to v7.0.0 — the v5.5.4 we were pinned to fails GPG verification since the codecovsecurity keybase account was deprecated. See https://github.com/codecov/codecov-action/issues/1956.

## Breaking changes

- Django 4.2 is no longer supported; consumers must run Django 5.2.
- Pact provider verification is no longer part of this repo. External systems firing the `contract_changed` `repository_dispatch` event will no-op.


ormsbee (commented):
🔥 🔥 🔥

ormsbee (approved):
🔥 🔥 🔥

